PR #136, opened 2014-06-23
Use Symbols as column names

### LINKED ISSUE #134: Should column names be symbols?
DataFrames.jl allows only symbol column names, whilst it looks like TimeSeries column names must be UTF8String.

https://github.com/JuliaStats/DataFrames.jl/pull/509

hayd:
fixes #134

Should work as before when passing strings as column names: they're cast as symbols.

New to Julia so comments/thoughts/pointers welcome.

*Now I think about it maybe should also check identifiers (again inline with DataFrames), but I don't think there is a Base implementation of that yet?...*

### DIFF
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -4,7 +4,7 @@
 for op in [:.+, :.-, :.*, :./]
   @eval begin
     function ($op){T,N}(ta1::TimeArray{T,N}, ta2::TimeArray{T,N})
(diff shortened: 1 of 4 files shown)


milktrader:
I'm closing this for now since the motivation behind using Symbols vs Strings for column names doesn't seem to be an improvement and will disallow spaces in column names.

hayd:
Apologies, I thought this was long closed. Thanks.

milktrader:
No problem. Thanks for tossing it out there for us to consider.

PR #141, opened 2014-06-30
add metadata field to TimeArray type closes #126

### LINKED ISSUE #126: Add meta element to TimeArray type
````julia
immutable TimeArray{T,N} <: AbstractTimeSeries

    timestamp::Vector{Date{ISOCalendar}}
    values::Array{T,N}
    colnames::Vector{ASCIIString}
    meta
end
````
There would be a default to leave this new field empty. A use case for this is a financial time series object. The meta field could be used to hold information about the underlying time series, such as Stock in terms of US Dollars.

milktrader:


milktrader:
@multidis here is where we left it with TimeSeries implementing a parameterized metadata field.

Three ways to go:

1. Do nothing
2. Add non-parameterized metadata field
3. Add parameterized metadata field

If we went with option 3 then there would need to be some rationale for adding this complexity. The main argument in favor is to be able to dispatch on it, so we need to ask ourselves if there is a method that would take advantage of it.

multidis:
The parameterized metadata dispatch becomes a fairly broad, not domain-specific question. [Performance checks](https://github.com/JuliaQuant/Roadmap/issues/2) should probably guide this choice, in line with `TimeSeries.jl` being more performance-oriented type as compared to e.g. [TimeData.jl](https://github.com/JuliaFinMetriX/TimeData.jl) which reminds me `xts` in R.

milktrader:
Issue #159 subsumes this.

PR #162, opened 2015-02-13
Param meta

milktrader:


milktrader:
Tests not passing on meta tests, specifically on v0.3 where apparently `Nothing != Void`

`Void` looks new but I missed the discussion about where it came from. (this can get metaphysical quick)

PR #174, opened 2015-03-03
Rtd

milktrader:


MrTomWhite:
Since everything in `/_build` Is generated from the make script, does it need to be committed to the repository?

milktrader:
I'm not sure and the answer is probably no. How would you handle this? Add it to .gitignore or just remove it before committing?

milktrader:
Hmm, I see that readthedocs can't build it the way it's presently laid out. It clearly doesn't care what's in the `_build` directory.

MrTomWhite:
I would probably do both so no one else makes the same commit again. :)

milktrader:
https://github.com/JuliaStats/TimeSeries.jl/issues/176

milktrader:
thanks @MrTomWhite

PR #181, opened 2015-04-26
Pr/180

milktrader:
I pulled https://github.com/JuliaStats/TimeSeries.jl/pull/180, corrected the path for the new tests


milktrader:
Fixing the problem behind this PR is something myself and @quinnj have put on the back burner. The kwarg where a format is specified should cover edge cases, but there needs to be a better parsing of possible combinations out there. I'm not really fond of how R has this hurdle to their odd time series conversions.

Here's the plan. 

Track down where in Date.jl datetime is parsed and simply update the regex engine being used there.

multidis:
Would be helpful to merge the kwarg option so that it becomes usable in the master package version. Oftentimes I have to do multiple steps to reformat timestamps the way `TimeSeries.jl` can take it and passing the format string would solve that.

Even if a more robust regex parsing is introduced in the future (which may also depend on Julia 0.4 Base-dates support modifications), it is always helpful to have an option for the user to specify the exact format instead of hoping for correct parsing.

milktrader:
Gotcha, good point. Not waiting for the perfect solution at the expense of a useful one. Let me get the documentation up to speed and we'll get this merged and tagged

milktrader:
@multidis I changed the `dtformat` to simply `format` with the bold assumption that users will understand that format is a date format. It's also documented.

milktrader:
fwiw, the zoo package in R uses the simple `format` kwarg.

milktrader:
gah, got this on nightlies

````
ERROR: LoadError: LoadError: LoadError: LoadError: LoadError: UndefVarError: Range1 not defined
````

milktrader:
I'm going to merge this but hold off on tagging it until I can figure out what happened to `Range1`. I tried a quick Compat workaround but it didn't work.

milktrader:
This has been included in tag `0.5.8` now. Needed to navigate a couple extra ERRORs.

PR #183, opened 2015-05-18
changed Range1 to UnitRange

milktrader:


milktrader:
gah, now this
````
Warning: both TimeSeries and Base export "values"; uses of it in module Main must be qualified
````
I used it first!

milktrader:
ooops, mucked up adding Base.values in the master branch so I'll just merge this before I become more confused.

PR #189, opened 2015-05-20
resolved deprecation warnings spit out in the test suite

milktrader:


milktrader:
`ERROR: FactCheck finished with 4 non-successful tests.`

Issue #158, opened 2015-01-25
Do we enforce timestamp to be unique and ordered?


tonyhffong:
from the implementation of `.+` etc, it doesn't seem to take advantage that it might be, nor do we have an efficient (binary search) test of the existence of a date/time in the TimeArray.

milktrader:
timestamp has to be unique and ordered per the inner constructor

````julia
# some initial conditions
 timestamp != unique(timestamp) ? error("there are duplicate dates"):
 ~(flipud(timestamp) == sort(timestamp) || timestamp == sort(timestamp)) ? error("dates are mangled"):
# some other checks then construction of object
````

**Comment by milktrader** [comment]
Not sure about your point regarding the `.+` operator though

**Comment by tonyhffong** [comment]
I'm referring to this code in apply.jl
```julia
for op in [:.+, :.-, :.*, :./]
  @eval begin
    function ($op){T,N}(ta1::TimeArray{T,N}, ta2::TimeArray{T,N})
      cname  = [ta1.colnames[1][1:2] *  string($op) *  ta2.colnames[1][1:2]]
      tstamp = Date[]
      vals   = T[]
      for i in 1:size(ta1.timestamp, 1)
        for j in 1:size(ta2.timestamp, 1)
          if ta1.timestamp[i] == ta2.timestamp[j] 
            push!(tstamp, ta1.timestamp[i]) 
            push!(vals, ($op)(ta1.values[i], ta2.values[j])) 
          end
        end
      end
      TimeArray(tstamp, vals, cname)
    end # function
  end # eval
end # loop
```
If the timestamps are already ordered, this could be a O(N) algorithm instead of a O(N^2) algorithm.

milktrader:
Needless to say I'd like to see that. I recall not finding a good solution to creating empty arrays and pushing into them.

GordStephen:
Borrowing heavily from the [`merge` implementation](https://github.com/JuliaStats/TimeSeries.jl/blob/master/src/combine.jl#L28), would something along these lines work?

```julia
for op in [:.+, :.-, :.*, :./]
  @eval begin
    function ($op){T,N}(ta1::TimeArray{T,N}, ta2::TimeArray{T,N})
      cname  = [ta1.colnames[1][1:2] *  string($op) *  ta2.colnames[1][1:2]]
      idx1, idx2 = overlaps(ta1.timestamp, ta2.timestamp)
      # obtain shared timestamp
[...]

PR #191, opened 2015-06-25
Switched to linear-time math and comparison operations between 1D Tim…

GordStephen:
O(n^2) -> O(n) TimeArray math and comparison speed-up as discussed in https://github.com/JuliaStats/TimeSeries.jl/issues/158. I also preallocated the output value array in the 1DArray  <-> comparison operations which should provide a minor performance enhancement as well.

Is there a particular reason .!= and ! aren't defined for TimeArray? I'm happy to add those?

### DIFF
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -4,6 +4,25 @@ COMPARE_DOTONLY = [:.>, :.<, :.==, :.>=, :.<=]
 
 ###### Mathematical operators  ###############
 
+function overlaps(t1, t2)
+    i = j = 1
+    idx1 = Int[]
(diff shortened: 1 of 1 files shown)


milktrader:
On the current master:

````julia
julia> using MarketData

julia> @elapsed BA["Close"] ./ AAPL["Close"]
25.19867983

julia> @elapsed BA["Close"] ./ AAPL["Close"]
25.173957679
````


On this PR:

````julia
using MarketData

julia> @elapsed BA["Close"] ./ AAPL["Close"]
0.028212047
````

Issue #194, opened 2015-06-26
Generalize value type from Array{T,N} to DenseArray{T,N}? 


GordStephen:
TimeArrays can't have BitArray contents at the moment. Moving allowed values up a step on the type hierarchy from `Array{T,N}` to `DenseArray{T,N}` would solve this (and also enable any other more obscure array types people might want to use) - thoughts?

milktrader:
I wanted to add speculative tag but can't figure out this api. In any case, thanks for the idea.

milktrader:
````julia
julia> Array <: DenseArray
true

julia> BitArray <: DenseArray
true

julia> BitArray <: Array
false
````

GordStephen:
Relevant discussion has moved over to https://github.com/JuliaStats/TimeSeries.jl/pull/197

PR #193, opened 2015-06-26
Added ! and .!=

GordStephen:
Added `!` and `.!=` TimeArray operations and relevant tests.

PR #195, opened 2015-06-27
Math/comparison/logic operation refactoring

GordStephen:
This refactors TimeArray math, comparison, and logic operations to be wrappers for base operations on the underlying value arrays - automatically aligning functionality with standard Array operations as defined by `Base` and saving some LOC.  Also adds symmetric functionality for broadcasting (2D __ 1D -> 2D, and 1D _ 2D -> 2D), elementwise `.^`, `.!=`, `%`, `.%`, unary math and logic operators (`!`, `+`, `-`) and binary logic operators (`&`, `|`, `$`), which allow things like:

```julia
julia> using MarketData

julia> (ohlc["Close"] .> 100) & (ohlc["Close"] .< 120)
500x1 TimeArray{Bool,1,DataType} 2000-01-03 to 2001-12-31

             Close.>100&Close.<120  
2000-01-03 | true                   
2000-01-04 | true                   
2000-01-05 | true                   
2000-01-06 | false                  
⋮
2001-12-26 | false                  
2001-12-27 | false                  
2001-12-28 | false                  
2001-12-31 | false
```
I took the liberty of changing the `TimeArray.values` type from `Array{T}` to `DenseArray{T}` (https://github.com/JuliaStats/TimeSeries.jl/issues/194) to accommodate BitArrays and implement the boolean operators... But I'm happy to go with a workaround if the preference is not to generalize the type.

Also, while `Base` implements bitwise operations on numbers and math operations on booleans, I thought that given the typical application context here that functionality would likely cause more bugs from user typos than provide benefits, and so left those out.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@
 * changed references of float(x) to map(Float64, x)
 * changed references of [a] to [a;] in a comprehension found in the by() method
 * added Compat package
-* substantial speedup for eleemnt-wise mathematical operators
+* substantial speedup for element-wise mathematical operators
 
 ### 0.5.9
 
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -1,128 +1,126 @@
-MATH_ALL        = [:.+, :.-, :.*, :./, :.^, :+, :-, :*, :/, :^]
-MATH_DOTONLY    = [:.+, :.-, :.*, :./]
-COMPARE_DOTONLY = [:.>, :.<, :.==, :.>=, :.<=] 
+UNARY         = [:+, :-, :!]
+MATH_NO_DOT   = [:+, :-, :*, :/, :^, :%]
+MATH_DOT      = [:.+, :.-, :.*, :./, :.^, :.%]
+COMPARISONS   = [:.>,:.<, :.==, :.>=, :.<=, :.!=]
+LOGICAL       = [:&, :|, :$]
 
-###### Mathematical operators  ###############
+SCALAR_OPS    = [MATH_NO_DOT; MATH_DOT; COMPARISONS]
+ELEMENT_OPS   = [MATH_DOT; COMPARISONS]
+BOOLEAN_OPS   = [LOGICAL; COMPARISONS]
 
-# TimeArray <--> TimeArray 
-for op in MATH_DOTONLY
+###### Math and Boolean Unary Operators  #####
+
+for op in UNARY
   @eval begin
-    function ($op){T,N}(ta1::TimeArray{T,N}, ta2::TimeArray{T,N})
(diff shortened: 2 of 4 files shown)


milktrader:
Whoa, that is quite the refactor!

I think it would be more digestible in smaller parts. Is there a significant performance improvement in using `basecall`?

GordStephen:
Only the new unary operators use `basecall` (since the other operations all take multiple inputs), but benchmarking against the original looping approach for `!` in https://github.com/JuliaStats/TimeSeries.jl/pull/193 doesn't show any significant difference (both 5-8ms on `AAPL`). The motivation for switching was to make the function clearer and more concise:

#193:
```julia
for op in UNARY
  @eval begin
    function ($op){T,N}(ta::TimeArray{T,N})
      cname  = [string($op) * ta.colnames[1][1:2] ]
      vals   = Array(Bool, length(ta))
      for i in 1:length(vals)
        vals[i] = ($op)(ta.values[i]) 
      end
      TimeArray(ta.timestamp, vals, cname, ta.meta)
    end # function
  end # eval
end # loop
```
vs #195 (this):
```julia
for op in UNARY
  @eval begin
    function ($op){T,N}(ta::TimeArray{T,N})
      cnames  = [string($op) * name for name in ta.colnames]
      basecall(ta, ($op), cnames=cnames) 
    end # function
  end # eval
end # loop
```

Looking at it now, the clearest and most consistent approach might be:
```julia
for op in UNARY
  @eval begin
    function ($op){T,N}(ta::TimeArray{T,N})
      cnames  = [string($op) * name for name in ta.colnames]
      vals = ($op)(ta.values)
      TimeArray(ta.timestamp, vals, cnames, ta.meta)
    end # function
  end # eval
end # loop
```
which would bring it in line with the other refactored operator implementations, for example
```julia
for op in SCALAR_OPS 
  @eval begin
    function ($op){T,N}(ta::TimeArray{T,N}, var::Union(BigInt,Signed,Unsigned,FloatingPoint))
      cnames  = [name * string($op) * string(var) for name in ta.colnames]
      vals = ($op)(ta.values, var)
      TimeArray(ta.timestamp, vals, cnames, ta.meta)
    end # function
  end # eval
end # loop
```

GordStephen:
Regarding it being more digestible in smaller parts, I can do a PR with my un-squashed development commits if you want?

milktrader:
How about a to do list and we can tackle one at a time? I'll file the issue.

milktrader:
Ok, here is the to do list started https://github.com/JuliaStats/TimeSeries.jl/issues/196

Closing this for now.

GordStephen:
Sounds good!

Issue #196, opened 2015-06-28
Refactor code base


milktrader:
- [ ] replace Arrays with DenseArrays
- [ ] add Unary operators
- [ ] add Logical operators
- [ ] refactor column naming convention
- [ ] add BigInt, Signed and Unsigned to dispatch list
- [ ] other

I'm stubbing out `other` for now and will fill it out with the _other_ changes @GordStephen is proposing

milktrader:
Besides the code, tests (which seem to be included) and documentation changes will need to be updated for the task to be marked complete.

GordStephen:
These ideas (or some evolution of them) all ended up in https://github.com/JuliaStats/TimeSeries.jl/pull/205, so I'll go ahead and close this.

PR #197, opened 2015-06-28
Switched to DenseArray and implemented unary and binary logic

GordStephen:
Points 1-3 in https://github.com/JuliaStats/TimeSeries.jl/issues/196:

- Replace Arrays with DenseArrays
- Add unary logic and math operators
- Add binary logic operators
- Add tests for new operators
- Update docs with logic operators

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,7 +7,7 @@
 * changed references of float(x) to map(Float64, x)
 * changed references of [a] to [a;] in a comprehension found in the by() method
 * added Compat package
-* substantial speedup for eleemnt-wise mathematical operators
+* substantial speedup for element-wise mathematical operators
 
 ### 0.5.9
 
--- a/docs/source/operators.rst
+++ b/docs/source/operators.rst
@@ -1,22 +1,23 @@
 Mathematical, comparison, and logical operators
 ===============================================
 
-TimeSeries supports common mathematical (such as ``+``), comparison ( such as ``==``)
+TimeSeries supports common mathematical (such as ``+``), comparison ( such as ``.==``)
 , and logic (such as ``&``) operators. The operations are only calculated on values that share a timestamp.
 
 mathematical
 ------------
 
 Mathematical operators create a TimeArray object where values are computed on shared timestamps when two TimeArray 
-objects are provided. Operations between a single TimeArray and ``Int`` or ``Float`` is also supported. The number 
-can precede the TimeArray object or vice versa (e.g. ``cl + 2`` or ``2 + cl``).
+objects are provided. Operations between a single TimeArray and ``Int`` or ``Float`` are also supported. The number
+can precede the TimeArray object or vice versa (e.g. ``cl + 2`` or ``2 + cl``). Broadcasting single-column arrays over multiple columns to perform operations is also supported.
 
-Except in the case of the ``/`` operator, both dot (``.+``) and non-dot (``+``) operations are supported. The semantics
+Except in the case of the ``/`` and ``^`` operators, both dot (``.+``) and non-dot (``+``) operations are supported when working with scalar values. The semantics
 of non-dot operations are fairly clear when working with time series data, where it is assumed that only equivalent 
 timestamped values are being operated on. 
 
-The exclusion of ``/`` from this logic is a special case. In matrix operations it has been confused with being 
-equivalent to the inverse, and because of the confusion base has excluded it. It is likewise excluded here. 
+The exclusion of ``/`` and ``^`` from this logic are special cases. In matrix operations ``/`` has been confused with being
+equivalent to the inverse, and because of the confusion base has excluded it. It is likewise excluded here. Base uses ``^`` to indicate matrix self-multiplication, and so it is not implemented
+ in this context.
 
 +------------------+------------------------------------------+
 | Operator         | Description                              |
@@ -29,7 +30,7 @@ equivalent to the inverse, and because of the confusion base has excluded it. It
 +------------------+------------------------------------------+
 |      ``./``      | mathematical element-wise division       |
 +------------------+------------------------------------------+
-| ``^`` or ``.^``  | mathematical element-wise exponentiation |
+|      ``.^``      | mathematical element-wise exponentiation |
 +------------------+------------------------------------------+
 | ``%`` or ``.%``  | mathematical element-wise remainder      |
 +------------------+------------------------------------------+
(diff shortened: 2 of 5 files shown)


milktrader:
You need DenseArrays for binary logic operators I presume?

GordStephen:
Yeah, `Base

GordStephen:
Yeah, the `Base` logic operations on arrays return BitArrays, so directly using those results in a TimeArray requires DenseArray support.

milktrader:
Why not use `Union{Array, BitArray}`?

milktrader:
Also, where did you find the type hierarchy? From experimentation?

GordStephen:
Yeah, for this, `Union{Array, BitArray}` would work just as well I think.

```julia
julia> super(Array)
DenseArray{T,N}

julia> super(BitArray)
DenseArray{Bool,N}

help?> DenseArray
DataType   : DenseArray{T,N}
  supertype: AbstractArray{T,N}
  subtypes : {Array{T,N},BitArray{N},SharedArray{T,N}}
```

milktrader:
Ok, good old REPL exploration. Why not go for the gold and use `AbstractArray`?

milktrader:
Probably goes to the issue of explication by elimination (see Jeff's thesis) in which case the more information you can provide the compiler, the better it is at optimizing. Before jumping over to `DenseArrays`, for example, it would be nice to get some benchmarks.

Here, btw, is `AbstractArray` subtypes:

````julia
help?> AbstractArray
search: AbstractArray AbstractSparseArray AbstractSparseMatrix

DataType   : AbstractArray{T,N}
  supertype: Any
  subtypes : Any[Base.LinAlg.AbstractTriangular{T,S<:AbstractArray{T,2}},Base.LinAlg.Bidiagonal{T},Base.LinAlg.Diagonal{T},Base.LinAlg.Hermitian{T,S<:AbstractArray{T,2}},Base.LinAlg.HessenbergQ{T,S<:AbstractArray{T,2}},Base.LinAlg.QRCompactWYQ{S,M<:AbstractArray{T,2}},Base.LinAlg.QRPackedQ{T,S<:AbstractArray{T,2}},Base.LinAlg.QRPackedWYQ{S,M<:AbstractArray{T,2}},Base.LinAlg.SVDOperator{T,S},Base.LinAlg.SymTridiagonal{T},Base.LinAlg.SymmetricRFP{T<:Union{Complex{Float64},Complex{Float32},Float64,Float32}},Base.LinAlg.Symmetric{T,S<:AbstractArray{T,2}},Base.LinAlg.TriangularRFP{T<:Union{Complex{Float64},Complex{Float32},Float64,Float32}},Base.LinAlg.Tridiagonal{T},Base.SparseMatrix.AbstractSparseArray{Tv,Ti,N},Base.SparseMatrix.CHOLMOD.FactorComponent{Tv,Ti,S},Core.Inference.Range{T},DenseArray{T,N},Range{T},SubArray{T,N,P<:AbstractArray{T,N},I<:Tuple{Vararg{Union{Range{Int64},Colon,Array{Int64,1},Int64}}},LD}]
````

**Comment by GordStephen** [comment]
In this context `DenseArray` seems to be the only `AbstractArray` subtype of relevance... I'll do some speed tests for the various levels of specificity though.

**Comment by milktrader** [comment]
Yep, agreed, and looking forward to the benchmarks

**Comment by GordStephen** [comment]
Seems like it's a wash, performance-wise:

`Array{T,N}` (`master`):
```julia
julia> using MarketData
julia> @elapsed AAPL["Close"] .+ BA["Close"]
0.043117468
```

`Union{Array{T,N},BitArray{N}}`:
```julia
julia> using MarketData
julia> @elapsed AAPL["Close"] .+ BA["Close"]
0.04327045
```

`DenseArray{T,N}` (this PR):
```julia
julia> using MarketData
julia> @elapsed AAPL["Close"] .+ BA["Close"]
0.04161412
```

`AbstractArray{T,N}`:
```julia
julia> using MarketData
julia> @elapsed AAPL["Close"] .+ BA["Close"]
0.04547258
```

Personally I'd lean towards `DenseArray` as a balance between generality and meaningful type restrictions.

milktrader:
Ok, thanks. In my mind it's `DenseArray` first, a work-around second and `Union{Array, BitArray}` last.

milktrader:
Just a copy-paste of the docs for reference
>DenseArray is an abstract subtype of AbstractArray intended to include all arrays that are laid out at regular offsets in memory, and which can therefore be passed to external C and Fortran functions expecting this memory layout. Subtypes should provide a method stride(A,k) that returns the “stride” of dimension k: increasing the index of dimension k by 1 should increase the index i of getindex(A,i) by stride(A,k). If a pointer conversion method convert(Ptr{T}, A) is provided, the memory layout should correspond in the same way to these strides.

The Array type is a specific instance of DenseArray where elements are stored in column-major order (see additional notes in Performance Tips). Vector and Matrix are aliases for the 1-d and 2-d cases. Specific operations such as scalar indexing, assignment, and a few other basic storage-specific operations are all that have to be implemented for :obj:`Array, so that the rest of the array library can be implemented in a generic manner.

GordStephen:
https://github.com/GordStephen/TimeSeries.jl/commit/3de4ebdca746637b1649d6aee5531a84ad9d020d  just adds a few new operators and relevants tests / docs - nothing too major. I'll probably push the refactoring elements of https://github.com/JuliaStats/TimeSeries.jl/pull/195 as another commit tomorrow.

milktrader:
Running this PR locally I'm not seeing any problems. My only concern is that there is some yet unknown effect. On the plus side, 

* performance is unaffected
* new operators are intoduced
* all tests are still passing

I need to grok the difference between `BitArray` and `Array{Bool, N}`.

milktrader:
ping @quinnj long story short is the replacement of the `values` element from `Array` to `DenseArray`, which provides support for `BitArray`

milktrader:
I've posted on google groups for a little feedback, but I think this implementation is probably the one to go with. https://groups.google.com/forum/#!topic/julia-dev/GwNwuwZflNI

milktrader:
@GordStephen there is some good feedback on the google groups post I did (link above). Apparently there is some question if `BitArray` is faster or slower than `Array{Bool,N}`. There is also some consensus building around using `AbstractArray` in any case, since using `DenseArray` has an expressed goal of being able to pass the object to BLAS, and that's not the case in `TimeArray`.

Maybe the least impact way to get the bitwise operators into the code is to have them return `Array{Bool.N}` for now and then we can discuss the merits of reworking the values component into an `AbstractArray`.

What do you think?

GordStephen:
I'm ok just to hold off on merging until the AbstractArray question has been settled - no point in implementing a workaround just to potentially change it again later (in the meantime, if anyone desperately needs bitwise operations, they can always pull in this fork themselves).

I don't see any mention of potential side-effects of switching to AbstractArray in that thread - as I understand it, generalizing would just allow TimeArray to wrap more types, but doesn't remove or change any existing functionality (but famous last words, maybe...)?

GordStephen:
Any new thoughts on this? Happy to address the merge conflicts if/when we choose to go ahead.

milktrader:
I say we go for this with the first tag that supports Julia 0.4, coming soon. Hopefully get to this today.

What did we decide was our best course here? `AbstratctArray` or `DenseArray`? I recall `AbstractArray` was the preferred choice.

GordStephen:
Yeah, I think the consensus settled on AbstractArray - the latest commit in this PR implemented that.

milktrader:
Ok, great. I'll tag up 0.5.11 (last version supporting Julia 03) and then 0.6.0 (first version supporting Julia 0.4 only) and then you can take it from there and tag 0.6.1 with the AbstractArray implementation. 

Remember to add a line to NEWS.md about the important improvement.

GordStephen:
Superseded by https://github.com/JuliaStats/TimeSeries.jl/pull/205

PR #201, opened 2015-09-04
Fix deprecation warnings

GordStephen:
Explicitly import math/inequality operators from Base to remove new 0.4 warnings like:

`WARNING: module TimeSeries should explicitly import .+ from Base`

A few tests had also started failing due a lack of agreeing significant digits with the FactCheck `roughly` function, e.g.

```
Failure :: (line:-1) :: correct alignment and operation between two TimeVectors :: fact was false
      Expression: cl .* op.values[1] --> roughly(11740.27)
        Expected: 11740.267199999998 ≅ 11740.27
```

so I increased the precision of the expected results.


GordStephen:
Oops, forgot that MarketData hasn't been fixed upstream for 0.4 (https://github.com/JuliaQuant/MarketData.jl/pull/18), so tests won't pass for that yet... Need to rework `import` for 0.3 support as well.

GordStephen:
Ok, 0.4 tests passing now too.

PR #204, opened 2015-09-13
first run of only Julia 0.4 support, tests passing locally

milktrader:


milktrader:
Okay, `iround` is deprecated in 0.4 and some other errors will make this a WIP

PR #205, opened 2015-09-13
Switch to AbstractArrays and refactor / generalize operator code

GordStephen:
A master-compatible version of https://github.com/JuliaStats/TimeSeries.jl/pull/197 - switches the value container from `Array` to `AbstractArray` and implements `!`, `~`, `&`, `|`, `$`, `%`, `!==`, and unary `+` and `-`, with appropriate documentation.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+### 0.6.1
+
+* generalized value container from Array to AbstractArray
+* implemented new element-wise operators: !, ~, &, |, $, %, !==
+* implemented element-wise unary math operators (+, -)
+
 ### 0.6.0
 
 * first version with support for Julia 0.4 only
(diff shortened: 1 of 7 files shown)


GordStephen:
Also, REQUIRE was still at `0.3 0.4-`, so I changed it - I'm assuming that's ok?

milktrader:
Nice catch @GordStephen , Tony found the same thing in METADATA so I just pushed version 0.6.1 to fix this.

milktrader:
Sorry I saw this after fixing 0.6.1 with the new REQUIRE. I think all that needs to be done is some fussing with the NEWS.md.

milktrader:
Let me take a stab at this right now.

milktrader:
Ok, phew. Done.

Thanks for the awesome work here @GordStephen !

milktrader:
I'm having some issues with tagging to METADATA that I'll sort out tomorrow. I'm trying to __fix__ 0.6.0 to reflect the correct julia versions, and then make this PR 0.6.1. Unfortunately I've already tagged 0.6.1 so I need to fix that too.

GordStephen:
Ok, sounds good!

PR #206, opened 2015-09-14
Indexing/lookup enhancements

GordStephen:
A bit of a grab bag of minor enhancements, generally indexing-related. Most notable is adding support for the `end` keyword in indices, and allowing specifying indices to get via a `TimeArray{Bool,1}`. To go along with that is a speedup for Date-based lookups, from polynomial to linear time:

Current master:
```julia
julia> @time AAPL[findwhen((BA["Close"] .> BA["Open"]) | (CAT["Close"] .> CAT["Open"]))]
  0.224620 seconds (730.40 k allocations: 42.386 MB, 5.90% gc time)
```
This PR:
```julia
julia> @time AAPL[(BA["Close"] .> BA["Open"]) | (CAT["Close"] .> CAT["Open"])]
  0.124473 seconds (668.16 k allocations: 41.726 MB, 7.25% gc time)
```

### DIFF
--- a/src/split.jl
+++ b/src/split.jl
@@ -27,31 +27,11 @@ end
 
 ###### findall ##################
 
-function findall(ta::TimeArray{Bool,1})
-    rownums = round(Int64, zeros(sum(ta.values)))
-    j = 1
-    for i in 1:length(ta)
-        if ta.values[i]
-            rownums[j] = i
-            j+=1
-        end
-    end
-    rownums
-end
- 
+findall(ta::TimeArray{Bool,1}) = find(ta.values)
+
 ###### findwhen #################
 
-function findwhen(ta::TimeArray{Bool,1})
-    tstamps = collect(Date(1,1,1):Year(1):Date(sum(ta.values),1,1))
-    j = 1
-    for i in 1:length(ta)
-        if ta.values[i]
-            tstamps[j] = ta.timestamp[i]
-            j+=1
-        end
-    end
-    tstamps
-end
+findwhen(ta::TimeArray{Bool,1}) = ta.timestamp[find(ta.values)]
 
 ###### element wrapers ###########
 
--- a/src/timearray.jl
+++ b/src/timearray.jl
@@ -1,6 +1,6 @@
 ###### type definition ##########
 
-import Base: convert, length, show, getindex, start, next, done, isempty
+import Base: convert, length, show, getindex, start, next, done, isempty, endof
 
 abstract AbstractTimeSeries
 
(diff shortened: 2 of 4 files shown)


src/split.jl
```diff
@@ -27,31 +27,11 @@ end
...
-    for i in 1:length(ta)
-        if ta.values[i]
-            rownums[j] = i
-            j+=1
-        end
-    end
-    rownums
-end
- 
+findall(ta::TimeArray{Bool,1}) = find(ta.values)
```
GordStephen: The `findall` function is really just a specific method of `Base.find` for TimeArrays. Would it make sense to embrace that and just call it `find`?

milktrader:
Yes, we can change the name. This will require us to follow a deprecation protocol.

milktrader:
https://github.com/JuliaStats/TimeSeries.jl/issues/208

milktrader:
Probably changing the name should be a separate PR

milktrader:
Good stuff, thanks!

milktrader:
If this is a stopping point for now, we can update NEWS.md (this will be 0.6.1 - more on this below) and tag it.

NOTE:

0.6.0 has been re-assigned to commit e3d5a679541bd4ed6052898667fcce656677f6d6 which includes the switch to AbstractArray. Release 0.6.1 has been removed and is available. I need METADATA to accept my PR for changing the sha1 and requires for the 0.6.0 tag and then we can move on.

GordStephen:
Yep, that's all for now.

milktrader:
Alright then, this should be tagged with these improvements. Can you add the notes to NEWS.md and then we'll stop committing to master until we can resolve the mess I created over at METADATA. Once that's resolved (the METADATA tagging), then we can tag this, publish it METADATA and begin work on your idea for using TimeType @GordStephen 

Also, I think it would be good to create a branch for the TimeType implementation.

````
git branch -b timetype
````
Push that to the repository and then we can merge it after it looks good to go.

**Comment by GordStephen** [comment]
Ok, I'll try and get to this tonight.

**Comment by milktrader** [comment]
Version 0.6.0 appears resolved at METADATA so you're good to go to add to NEWS and then tag version 0.6.1.
````
# after committing NEWS
julia> Pkg.update()
julia> Pkg.tag("TimeSeries", :patch)
julia> Pkg.publish() # this will open a PR window in METADATA's repo
````

**Comment by GordStephen** [comment]
So on step 2 I get:

```julia
julia> Pkg.tag("TimeSeries", :patch)
INFO: Tagging TimeSeries v0.6.1
fatal: tag 'v0.6.1' already exists
```

Any ideas?

milktrader:
Yeah, I was afraid of that. Since you pushed the changes, probably the easiest way is to remove the repo locally and then clone it again. You can see (here)[https://github.com/JuliaStats/TimeSeries.jl/releases] that there is no 0.6.1

milktrader:
Well now I'm getting this problem with phantom 0.6.1 ... gah

milktrader:
Here's what I propose.

I push to METADATA 0.6.1 which should be harmless. Then we do a Pkg.update() and hopefully we get a clean slate.

milktrader:
@GordStephen let me clean up this mess. I'll fix the NEWS and get a new tag  on this PR (should be 0.6.2).

milktrader:
Alright then, it looks like things should be back on track. Just waiting to get merged into METADATA. 

Sorry for all this noise!

Issue #207, opened 2015-09-14
Type parameter for Date vs DateTime?


GordStephen:
Thinking about how to approach the remaining issues in #190, it occurs to me that it would be useful to be able to define methods specific to Date- vs DateTime-indexed arrays. One approach might be to specify the TimeType as a TimeArray type parameter. The type definition might look something like:

```julia
immutable TimeArray{T,N,D<:TimeType,M} <: AbstractTimeSeries

    timestamp::Vector{D}
    values::AbstractArray{T,N}
    colnames::Vector{UTF8String}
    meta::M
```
One could then write things like:

```julia
function from{T,N,D}(ta::TimeArray{T,N,D}, starttime::D)
...
```
Any thoughts?

milktrader:
I presume that the Dates package in Base has an abstract TimeType that parents Date and DateTime.

@quinnj ?

GordStephen:
Yep, correct.

```julia
julia> subtypes(Dates.TimeType)
2-element Array{Any,1}:
 Base.Dates.Date    
 Base.Dates.DateTime
```

milktrader:
Interesting. Starting to think about your interface above. It looks elegant and presumably solves the issue about supporting both Date and DateTime operations, no?

GordStephen:
Yeah, exactly. It would also theoretically support any new or custom TimeTypes that were to come along in the future.

milktrader:
Good point, like a `Ticktime` type maybe for high frequency trading

GordStephen:
Added in #223

PR #211, opened 2015-09-26
WIP: Adding TimeType type parameter

GordStephen:
As discussed in https://github.com/JuliaStats/TimeSeries.jl/issues/207, this would add a type parameter to `TimeArray` describing the type of the timestamps being used.

One question raised by the failing tests is, what to do with index lookups when the timestamp index passed in isn't the same type as the timestamps? Historically Date and DateTime have been used a bit interchangeably, but if we want to support other theoretical TimeTypes this may have to be revisited...

Also, the tests (for both master and this branch) started failing on me all of a sudden without `using` TimeSeries and Base.Dates in `runtests.jl`. I'm a bit puzzled because Travis didn't seem to have a problem with the same commits, and as far as I could tell was running the same versions of Julia, FactCheck, etc... Not sure what was up there, but I added in the necessary lines here.

### DIFF
--- a/src/combine.jl
+++ b/src/combine.jl
@@ -29,7 +29,7 @@ end
 
 # collapse ######################
 
-function collapse{T,N}(ta::TimeArray{T,N}, f::Function; period::Function=week)
+function collapse{T,N,D}(ta::TimeArray{T,N,D}, f::Function; period::Function=week)
   
     w = [period(ta.timestamp[t]) for t in 1:length(ta)] # get weekly id from entire array
     z = Int[]; j = 1
@@ -48,7 +48,7 @@ function collapse{T,N}(ta::TimeArray{T,N}, f::Function; period::Function=week)
     push!(z, z[length(z)] + 1)  
    
     # pre-allocate timestamp and value arrays
-    tstamps = [(Date(1,1,1):Year(1):Date(maximum(z),1,1));]
+    tstamps = Array{D}(maximum(z))
     vals    = zeros(maximum(z)) # number of unique periods
     #replace their values except for the last row 
     for i = 1:maximum(z)-1  # iterate over period ID groupings
--- a/src/split.jl
+++ b/src/split.jl
@@ -17,15 +17,15 @@ end
  
 # from, to ######################
  
-function from{T,N,D}(ta::TimeArray{T,N,D}, d::D)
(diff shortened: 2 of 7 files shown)


GordStephen:
@milktrader are you OK if we restrict lookups to the TimeType of the array timestamps? Right now it's possible to acess a Date-indexed row with a DateTime object, for example.

GordStephen:
Changed to/from/collapse to be TimeType-agnostic. Tests currently fail due to breaking changes this PR would cause:

1 - Loss of ability to lookup Date-indexed rows by a DateTime object
2 - `to` and `from` methods change from:

```julia
to{T,N}(ta::TimeArray{T,N}, y::Int, m::Int, d::Int)
```

```julia
to{T,N,D}(ta::TimeArray{T,N,D}, d::D)
```

If everyone's ok with these, we can add deprecation warnings to the next release and then make the changes later on, and I'll update the tests here accordingly.

milktrader:
Sorry for being out the last couple weeks, Ironman is a time sink if you ever decide to don one (pro tip). 

I have to think about this PR a little.

milktrader:
Ok, I scanned over the code and put down some notes. Overall, I like this idea. Since this is a significant change, we'll probably want to go to version 0.7.0 with these changes. (this would be something to add to the NOTES.md). 

You mentioned 
>Changed to/from/collapse to be TimeType-agnostic. Tests currently fail due to breaking changes this >PR would cause:

>1 - Loss of ability to lookup Date-indexed rows by a DateTime object
>2 - to and from methods change from:

How about his approach ... you complete the transition to TimeType and push it as a branch. Call it 0.7.0, for example. We can then experiment with it some more there and merge it with master after running it through some more paces. 

Version 0.6.x is fairly stable now and I expect that even with some minor changes to it, merging this big change will only create a manageable amount of merge conflicts that should be straightforward to correct.

GordStephen:
Ok, sounds good. We should probably tag at least one more 0.6.x version that adds warnings to methods that 0.7.0 would deprecate (current to/from methods, etc).

milktrader:
Go ahead and merge and tag a new 0.6.x version (:patch) and then we can stabilize the 0.6.x (master) branch for work on the 0.7.0 branch.

GordStephen:
In the process of being implemented in #223

PR #215, opened 2015-10-16
WIP: Merge extensions

GordStephen:
Eventually it would be good to have left, right, and outer merges/joins as well as the default inner merge - maybe once we have a better means of handling missing values though. I needed to induce equal spacing of observation rows and figured a left merge would be the cleanest way to do that, so implemented it and leave it here as a potential starting point if we want to come back to this somewhere down the road.

For reference, here's what the uniform spacing functions do:

```julia
julia> using TimeSeries
julia> using MarketData
julia> uniformlyspaced(cl)
false
julia> uniformlyspace(cl)[1:8]
8x1 TimeSeries.TimeArray{Float64,2,Void} 2000-01-03 to 2000-01-10

             Close     
2000-01-03 | 111.94    
2000-01-04 | 102.5     
2000-01-05 | 104.0     
2000-01-06 | 95.0      
⋮
2000-01-07 | 99.5      
2000-01-08 | NaN       
2000-01-09 | NaN       
2000-01-10 | 97.75
```

### DIFF
--- a/src/TimeSeries.jl
+++ b/src/TimeSeries.jl
@@ -9,8 +9,8 @@ using Base.Dates
 export TimeArray, AbstractTimeSeries,
        by, from, to, findwhen, findall, timestamp, values, colnames, meta,
        lag, lead, percentchange, moving, upto,
-       basecall,
-       merge, collapse,
+       basecall, uniformlyspaced, uniformlyspace,
+       merge, merge_left, collapse,
        readtimearray
 
 ###### include ##################
--- a/src/combine.jl
+++ b/src/combine.jl
@@ -2,11 +2,9 @@ import Base: merge
 
 ###### merge ####################
 
-# used by merge method
-
 # thanks Tom Short @tshort for this implementation
 
-function merge{T}(ta1::TimeArray{T}, ta2::TimeArray{T}; col_names=[""])
+function merge{T}(ta1::TimeArray{T}, ta2::TimeArray{T}; col_names::Vector=[])
     # first test metadata matches
     ta1.meta == ta2.meta ? meta = ta1.meta : error("metadata doesn't match")
     # obtain unique indexes of when dates match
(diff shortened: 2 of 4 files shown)


src/combine.jl
```diff
@@ -21,12 +19,33 @@ function merge{T}(ta1::TimeArray{T}, ta2::TimeArray{T}; col_names=[""])
     # combine existing colnames
     cnames = vcat(ta1.colnames, ta2.colnames)
     # check if kwarg to over-ride simple vcat and then if colnames is valid length
-    size(col_names,1) == 1 ? cnames = cnames :       # kwarg not supplied
-    size(col_names,1) == size(vals,2) ? cnames = col_names : error("col_names supplied is not correct size")
+    length(col_names) == size(vals,2) ? cnames = col_names :
```
milktrader: Why change `size(col_names,1)` to `length(col_names)`? I'd prefer to not use `length`

src/TimeSeries.jl
```diff
@@ -9,8 +9,8 @@ using Base.Dates
 export TimeArray, AbstractTimeSeries,
        by, from, to, findwhen, findall, timestamp, values, colnames, meta,
        lag, lead, percentchange, moving, upto,
-       basecall,
-       merge, collapse,
+       basecall, uniformlyspaced, uniformlyspace,
+       merge, merge_left, collapse,
```
milktrader: Can we keep only `merge` and make `merge_left` available with a kwarg, setting it to false?

milktrader:
I like where you're going with this. Maybe this can be part of version 0.7.0 branch? TimeType and supporting this functionality are a big (and I think positive) changes to the package

src/TimeSeries.jl
GordStephen: Definitely, this was just a quick-and-dirty first pass - ultimately I think we should have a single merge function with various methods for inner, left, right, and outer joins.

src/combine.jl
GordStephen: Given the new `Vector` type contraint for `col_names`, I  thought `length(col_names)` would be clearer than `size(col_names, 1)` while remaining unambiguous - I can certainly live with the latter option though if you'd prefer to keep it.

src/TimeSeries.jl
milktrader: ok, good. Another argument in favor of *experimenting* in the 0.7.0 branch

src/combine.jl
milktrader: Ok, this is a valid argument in favor of using `length`

GordStephen:
Incorporated into https://github.com/JuliaStats/TimeSeries.jl/commit/c88c3c44900126094383684abe6a435aa0bcbed9 in the 0.7.0 branch.

PR #219, opened 2015-10-25
Fix BoundsError when calling show on empty TimeArray

GordStephen:
Also adds tests for `show`ing various lengths of TimeArray.

Example:

```julia
julia> using MarketData

julia> ohlc[1:0]
0x4 TimeSeries.TimeArray{Float64,2,Void}

   Open   High   Low    Close  

```

### DIFF
--- a/src/timearray.jl
+++ b/src/timearray.jl
@@ -68,17 +68,18 @@ function show{T,N}(io::IO, ta::TimeArray{T,N})
             intcatcher[c] = true
         end
     end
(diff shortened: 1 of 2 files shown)


GordStephen:
Most of the remaining untested code was in `show`, so this provides a big coverage boost as well!

GordStephen:
If no one has any issues with this, I'll merge at the end of the week.

milktrader:
:+1: 

Don't forget to update README and tag a minor release.

Issue #218, opened 2015-10-19
Proposed changes coming out in Version 0.7.0


milktrader:
These are some important changes coming, mostly being developed by @GordStephen 

- [x] replace `Union{Date, DateTime}` with `TimeType` in the timestamp
- [x] improve the `merge` method to support more types of merges
- [x] add `NaN` as a missing, consumed value sentinel option until NullableArrays matures
- [x] streamline `lead` / `lag` API (#221)
- [x] rename `findall` to `find`(https://github.com/JuliaStats/TimeSeries.jl/pull/206#discussion_r39359627)
- [x] remove `meta` type parameter (#220) 

I would say that we keep with only supporting Julia 0.4.x with this version, and move to support Julia 0.5.x with TImeSeries 0.8.0

milktrader:
@GordStephen you can of course edit the list above and add or modify entries.

GordStephen:
Sounds good, will do.

GordStephen:
All of the above now addressed in #223

PR #223, opened 2015-10-26
WIP: Version 0.7.0

GordStephen:
Implements changes discussed in https://github.com/JuliaStats/TimeSeries.jl/issues/218


GordStephen:
Thought: it would be good to be able to specify the missingness value when non-inner merging (e.g. default to `NaN`, but can be overridden to be `0`, etc)

milktrader:
Can you provide an example. In general, I'd be against using `0` as a sentinel for missing.

GordStephen:
In general I would be too, but if for example you wanted to supplement an existing TimeArray `ta1` with columns with some default value to be overridden by specific rows from another TimeArray `ta2` when possible, it would be nice to be able to do a `merge(ta1, ta2, Val{:left}, defaultval=x)` that filled the new columns with either the specified default values `x` (could be `0`, `Inf`, `false`, or whatever other value the situation calls for), or the values contained in `ta2`. The default `defaultval` would remain `NaN` and would only change if explicitly specified.

milktrader:
Okay, this seems reasonable. This way you can only trick yourself by explicitly passing in a bogus value. 

The case where this is a gotcha is if you're taking a mean of values and all of a sudden you have a block of zeroes contributing to the calculation where the real value is unknown or simply doesn't exist.

andreasnoack:
Sorry or entering the discussion here so late. I've been following the discussion, but I haven't really had time to participate. Some comments

I think it's a very good idea to have a type parameter for dates instead of using the union. As you've already mentioned, this makes it possible to use a different date implementation later, but an important reason is also the costs for accessing a field with a non-concrete type such as a union. This is also the case for the `values` field, so I'll propose either to make that field parametric on the array type, in the same way as `SubArray` or `LowerTriangular`, or simply restrict it to `Matrix`. I wouldn't expect sparse matrices to be so common here, but I might be wrong. As an example of the costs see this example:
```jl
julia> type A{T}
       data::AbstractMatrix{T}
       end

julia> type B{T,S<:AbstractMatrix}
       data::S
       end


julia> let
       tmpA = A(randn(10,10))
       tmpB = B{Float64,Matrix{Float64}}(randn(10,10))
       @time for i = 1:10^7; tmpA.data[1,1]; end
       @time for i = 1:10^7; tmpB.data[1,1]; end
       end
  0.270920 seconds (10.00 M allocations: 152.588 MB, 6.59% gc time)
  0.004150 seconds
```

Just a side question: why not merge the sub items of #218 as they are completed instead of bundling them all together here?

GordStephen:
Interesting point about parametrizing on the array type. Is there a way to do that while both maintaining support for wrapping things like `BitArray`s and ideally keeping the array value and dimension type parameters? Something along these lines (if it were legal)

```julia
immutable TimeArray{T, N, D <: DateTime, S <: AbstractArray{T,N}}
    timestamp::Vector{D}
    values::S
    colnames::Vector{UTF8String}
    meta::Any

...

end
```

Back when we switched from `Array{T,N}` to `AbstractArray{T,N}` we did some performance tests and didn't see a major performance hit in moving to the abstract type - I wonder if we missed something, or if the math operations just dominate the execution time and marginalize the performance benefits? https://github.com/JuliaStats/TimeSeries.jl/pull/197#issuecomment-116303398

GordStephen:
Some performance tests - more for my own understanding than anything: https://gist.github.com/GordStephen/209eac0b2a0a1bee4215

andreasnoack:
@GordStephen We are many who'd like to be able to restrict `S` in the way you discribe. However, I've been told that it is complicated to implement, but it might come. For now, the restriction will have to be enforced through the constructors and that is indeed how `SubArray` and Lower/UpperTriangular` try to get around this limitation.

I think the timings in #197 are simply dominated by the operations after the load as you also suggest. The price is paid each time you access the field. If you start out by loading the underlying array, e.g.
```jl
B = A.values
```
and compute on `B`, then you only pay the price once, but that is probably not always possible or a natural way of writing code.

GordStephen:
Ok, just tied up a bunch of loose ends (discussed here and in the issues referenced above) - I think we're good to go with this. I'll merge tomorrow morning unless there are any last-minute concerns.

milktrader:
Merge away. Let's take it through the paces before we tag it as 0.7.0 though.

GordStephen:
Ok, done.

milktrader:
Great work! Looking over it and updating NEWS shortly ...

milktrader:
I've just pushed the NEWS.md. If you don't mind checking that I didn't mis-state or omit anything @GordStephen

GordStephen:
Looks good to me!

milktrader:
Alright then, I'll tag this and push it up to METADATA.

PR #234, opened 2015-11-24
introduce when method and begin deprecating by method

### LINKED ISSUE #231: Refactor "by" method
- [x] make new `when` method period argument non-kwarg with a default value
- [x] re-order argument list in new `when` method to `TimeArray`,  `Function`, `Int or String`
- [x] allow interval argument to take a string (for `dayname` and `monthname`) -- both `by` and `when`
- [x]  write tests to cover all possible arguments
- [x] update NEWS
- [x] create deprecated.jl `src` and `test` files
- [x] create deprecation warnings
- [ ] ~~possibly not name the new method `when` but overload `Base.find` instead.~~ https://github.com/JuliaStats/TimeSeries.jl/issues/232

milktrader:
Closes #231


milktrader:
@GordStephen when you get a chance, would you mind looking this over. If you're fine with it I'll merge it and tag a new release.

milktrader:
Thanks for the feedback @GordStephen 

I'll merge this, tag it and then maybe we can focus on your 0.7.0 branch for the next tag ...

PR #236, opened 2015-12-15
show NaN sentinels as NA

milktrader:
I'm tempted to further define missing values as `CBC` (consumed by calculation), `M` (missing), `S` (shifted) and so forth. 

This looks better in my opinion 

````julia
julia> using MarketData, TimeSeries

julia> lag(cl, padding=true)
500x1 TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}} 2000-01-03 to 2001-12-31

             Close     
2000-01-03 | NA        
2000-01-04 | 111.94    
2000-01-05 | 102.5     
2000-01-06 | 104.0     
⋮
2001-12-26 | 21.36     
2001-12-27 | 21.49     
2001-12-28 | 22.07     
2001-12-31 | 22.43     
````

### DIFF
--- a/src/timearray.jl
+++ b/src/timearray.jl
@@ -101,6 +101,8 @@ function show{T,N}(io::IO, ta::TimeArray{T,N})
             print(io, rpad(ta.values[i,j], colwidth[j] + 2, " ")) :
             intcatcher[j] & SHOWINT ?
             print(io, rpad(round(Integer, ta.values[i,j]), colwidth[j] + 2, " ")) : 
+            isnan(ta.values[i,j]) ?
+            print(io, rpad("NA", colwidth[j] + 2, " ")) :
             print(io, rpad(round(ta.values[i,j], DECIMALS), colwidth[j] + 2, " "))
         end
         println(io, "")
@@ -115,6 +117,8 @@ function show{T,N}(io::IO, ta::TimeArray{T,N})
             print(io, rpad(ta.values[i,j], colwidth[j] + 2, " ")) :
             intcatcher[j] & SHOWINT ?
             print(io, rpad(round(Integer, ta.values[i,j]), colwidth[j] + 2, " ")) : 
+            isnan(ta.values[i,j]) ?
+            print(io, rpad("NA", colwidth[j] + 2, " ")) :
             print(io, rpad(round(ta.values[i,j], DECIMALS), colwidth[j] + 2, " "))
         end
         println(io, "")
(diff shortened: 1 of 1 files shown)


nalimilan:
Rather than `NA`, it would make sense to show these as `#NULL` as in NullableArrays. The goal is to have a common interface, right (https://github.com/JuliaStats/NullableArrays.jl/issues/93)?

milktrader:
Yes, in favor of a common interface. Thanks for commenting and linking, I don't see this PR as a settled solution but rather a proposal. 

I'm not sure we can't be more specific than `#NULL` (which is difficult to look at), `NaN` (which is misleading) or `NA` (which has been the traditional compromise in languages such as R).

There is an outstanding issue to implement Nullable in this package, but @ViralBShah warned me at JuliaCon this year that there are some performance issues that need to be resolved. I know that the NullableArrays.jl package aims to address these.

GordStephen:
Interesting idea to go beyond just `NaN` or `#NULL`... I assume consumed-by-calculation would relate to things like differencing and shifted would result from lead/lag operations? Are you thinking that there are situations where knowing the cause of the missingness could allow for more effective use of the data further down a processing pipeline?

milktrader:
__Consumed by calculation__ would have moving average as its canonical use case. To take the 10-period moving average, you need to _consume_ the first nine values. 

By placing `CBC` in that slot (and I'm only proposing to do this by _showing_) you can keep track of why there are missing values in a certain time stamp.

Shifting (which is lag and lead consequences) could be represented differently. Maybe with unicode arrows representing the magnitude of the period shift? ↓ or 3↓

milktrader:
To start a short list of the idea ...

* consumed-by-calculation (moving average) `CBC`
* shifted (lag) ↓
* exists but is not known (missing data point - FRED data uses a `.` here) `missing` ?
* does not exist (stock market closed on Christmas, which is a Tuesday, e.g.) `DNE`?

Programmatically, why not just pop in the `NaN` float in these slots and be done with it?

The showing would merely add meaningful information to this slot.

GordStephen:
Hmm, interesting. Maybe this takes the form of a general-purpose / user-extendable display output masking feature that built-in functions can make specific use of? The TimeArray type would get a new `Array{DisplayMask}` field with `DisplayMask(::AbstractString, ::BitArray{N}` (or the option for a sparse BitArray might be useful too)?

milktrader:
This is an interesting implementation idea. So far I've just been spitballing but `DisplayMask` may be a way to implement it. Wondering about the performance hit though.

milktrader:
This has gotten stale and I've been thinking this is better suited for a separate package.

PR #241, opened 2015-12-29
Implement vcat and map for TimeArray

dourouc05:
Usual functions to work with arrays. Implemented `vcat` and not `hcat` or `cat`, because they sound more like `merge`.


milktrader:
These look like reasonable adds, what do you think @GordStephen ?

src/combine.jl
```diff
@@ -82,3 +82,41 @@ function collapse{T,N,D}(ta::TimeArray{T,N,D}, f::Function; period::Function=wee
...
+  # Concatenate the contents. 
+  timestamps = vcat([ta.timestamp for ta in TA]...)
+  values = vcat([ta.values for ta in TA]...)
+  return TimeArray(timestamps, values, TA[1].colnames, TA[1].meta)
+end
+
+# map ######################
+
+function map{T,N,D,A}(f::Function, ta::TimeArray{T,N,D,A})
+  timestamps = Array(typeof(ta.timestamp[1]), length(ta))
```
GordStephen: Can this just be `timestamps = similar(ta.timestamp)`?

GordStephen:
Looks good to me!

milktrader:
Ok, we can clean up the indentation later. Thanks @dourouc05

PR #244, opened 2016-01-09
Multi-column collapse

### LINKED ISSUE #243: Is collapse() supposed to work with multiple columns?
Using `collapse()` with multiple columns fails with the following error message

```julia
julia> collapse(ta, sum)
ERROR: column names must match width of array
 in call at /Users/abisen/.julia/v0.4/TimeSeries/src/timearray.jl:19
 in collapse at /Users/abisen/.julia/v0.4/TimeSeries/src/combine.jl:83
```

This is how the TimeArray is constructed

```julia
julia> dates=collect(Date(2000,1,1):Date(2000,12,31))
julia> ta=TimeArray(dates, [rand(length(dates)) rand(length(dates)).*100], ["a", "b"])
```

When using a single column everything works as it's supposed to

```julia
julia> collapse(ta["a"], sum, period=month)
12x1 TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}} 2000-01-31 to 2000-12-31

             a         
2000-01-31 | 13.1665   
2000-02-29 | 12.3132   
2000-03-31 | 19.1026   
2000-04-30 | 14.4827   
⋮
2000-09-30 | 13.2207   
2000-10-31 | 14.3403   
2000-11-30 | 14.3449   
2000-12-31 | 186.1699  
```

but when using both the columns it errors out

```julia
julia> collapse(ta, sum, period=month)
ERROR: column names must match width of array
 in call at /Users/abisen/.julia/v0.4/TimeSeries/src/timearray.jl:19
 in collapse at /Users/abisen/.julia/v0.4/TimeSeries/src/combine.jl:83
```

GordStephen:
Fixes #243 to allow `collapse` operations on 2D TimeArrays, with appropriate tests. Scores a nice speedup as well:

`master`:
```julia
julia> using TimeSeries, MarketData

julia> collapse(BA["Open"], last); @time collapse(BA["Open"], last);
  0.188057 seconds (294.44 k allocations: 46.848 MB, 2.21% gc time)
```

this PR:
```julia
julia> using TimeSeries, MarketData

julia> collapse(BA["Open"], last); @time collapse(BA["Open"], last);
  0.026178 seconds (118.08 k allocations: 33.954 MB, 12.61% gc time)
```

While we're looking at this, should there be an option for what timestamp is used to represent the downsampled rows? The current approach is to use the last timestamp in the cluster, although I think using the first timestamp would be just as justifiable (in fact preferable for my typical use cases).

Also, I think ideally the collapsed timestamps would represent the beginning/end of the specified time period, not just the first/last timestamp observed in said period - for example if I'm collapsing data down to a monthly average but my first observation in January is 11pm on the 3rd, I'd rather have the average associated with `2016-01-01T00:00:00` as opposed to `2016-01-03T23:00:00`. Implementing this would be more involved though, and would probably be hard to keep generalized for arbitrary `TimeType` timestamps...

### DIFF
--- a/docs/source/combine.rst
+++ b/docs/source/combine.rst
@@ -61,8 +61,7 @@ collapse
 The ``collapse`` method allows for compressing data into a larger time frame. For example, converting daily data into monthly data.
 When compressing dates, something rational has to be done with the values that lived in the more granular time frame. To define what
 happens, a function call is made. In our example, we want to compress the daily ``cl`` closing prices from daily to monthly. It makes
-sense for us to take the last value known and have that represented with the derived timestamp. The ``collapse`` keyword argument for
-time frame is the ``week`` method, which is defined in ``Base.Dates``. A list of valid time methods is presented below.
+sense for us to take the ``last`` value known and have that represented with the corresponding timestamp. A non-exhaustive list of valid time methods is presented below.
 
 +--------------+-------------+
 | Dates method | Time length |
@@ -78,7 +77,7 @@ time frame is the ``week`` method, which is defined in ``Base.Dates``. A list of
 
 Showing this code in REPL::
 
-    julia> collapse(cl,last,period=month)
+    julia> collapse(cl,month,last)
     24x1 TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}} 2000-01-31 to 2001-12-31
 
                  Close
@@ -91,6 +90,23 @@ Showing this code in REPL::
     2001-10-31 | 17.56
     2001-11-30 | 21.3
     2001-12-31 | 21.9
+
+We can also supply the function that chooses the timestamp and the function that determines the corresponding value independently::
+
+    julia> collapse(cl, month, last, mean)
+    24x1 TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}} 2000-01-31 to 2001-12-31
+
+		Close     
+    2000-01-31 | 103.3595  
+    2000-02-29 | 111.6375  
+    2000-03-31 | 128.5026  
+    2000-04-28 | 123.1058  
+    ⋮
+    2001-09-28 | 16.602    
+    2001-10-31 | 17.3222   
+    2001-11-30 | 19.649    
+    2001-12-31 | 21.695    
+
 
 vcat
 ----
--- a/src/combine.jl
+++ b/src/combine.jl
@@ -44,7 +44,7 @@ end
 
 # collapse ######################
 
-function collapse{T,N,D}(ta::TimeArray{T,N,D}, value::Function; period::Function=week, timestamp::Function=last)
+function collapse{T,N,D}(ta::TimeArray{T,N,D}, period::Function=week, timestamp::Function=last, value::Function=timestamp)
 
     length(ta) == 0 && return ta
 
--- a/src/combine.jl
+++ b/src/combine.jl
@@ -63,7 +63,7 @@ function collapse{T,N,D}(ta::TimeArray{T,N,D}, f::Function; period::Function=wee
 
         if mapped_tstamp != next_mapped_tstamp
           push!(collapsed_tstamps, tstamp)
-          collapsed_values = [collapsed_values; [f(ta.values[cluster_startrow:i, j]) for j in 1:ncols]']
+          collapsed_values = [collapsed_values; T[f(ta.values[cluster_startrow:i, j]) for j in 1:ncols]']
           cluster_startrow = i+1
         end #if
 
@@ -73,7 +73,7 @@ function collapse{T,N,D}(ta::TimeArray{T,N,D}, f::Function; period::Function=wee
     end #for
 
     push!(collapsed_tstamps, tstamp)
-    collapsed_values = [collapsed_values; [f(ta.values[cluster_startrow:end, j]) for j in 1:ncols]']
+    collapsed_values = [collapsed_values; T[f(ta.values[cluster_startrow:end, j]) for j in 1:ncols]']
 
     N == 1 && (collapsed_values = vec(collapsed_values))
     return TimeArray(collapsed_tstamps, collapsed_values, ta.colnames, ta.meta)
--- a/src/combine.jl
+++ b/src/combine.jl
@@ -44,7 +44,7 @@ end
 
 # collapse ######################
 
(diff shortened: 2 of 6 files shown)


GordStephen:
How about if the timestamp selector works just like `f` and is a function mapping the clustered groups of timestamps (rather than values) to a single value? So one could use `first` or `last` or define their own function. That way you could write custom functions to return `2016-01-01T00:00:00` given a list of datetimes in January, etc.

GordStephen:
Implemented that - the new keyword argument name is rather ungainly right now, but it's starting point:

```julia
function collapse{T,N,D}(ta::TimeArray{T,N,D}, collapse_valuecluster::Function; period::Function=week, collapse_timestampcluster::Function=last)
```

Regarding what's a mandatory argument and what has defaults, making `period` mandatory and making `collapse_timestampcluster` and `collapse_valuecluster` optional arguments (defaulting to something like first/first or first/mean or last/last) seems more natural to me, but I know everyone has different use cases.

milktrader:
I'm wondering if we can compress the verbose collapse_timestampcluster, collapse_valuecluster arguments.

milktrader:
Why not just `value` and `timestamp`?

GordStephen:
Sure, sounds good to me. What do you think about changing the mandatory/optional arguments around to either

```julia
collapse{T,N,D}(ta::TimeArray{T,N,D}, period::Function, value::Function=first, timestamp::Function=first)
```
or
```julia
collapse{T,N,D}(ta::TimeArray{T,N,D}, period::Function, value::Function=last, timestamp::Function=last)
```
`period` doesn't seem to have a natural default value, especially if we want to support arbitrary TimeTypes, but there are choices for `value` and `timestamp` that seem like intuitive context-agnostic defaults. Of course, we'd need to tread carefully (or change the function name) since it would be a breaking change.

milktrader:
For financial time series, the natural defaults would be `first` timestamp and value `last`. What use case do you have for the other time series.

GordStephen:
I would typically be using `first` for timestamps and probably end up explicitly passing in `mean` or `sum` for values in most cases (power flows or energy consumption), so if no one else has objections to `last` as a type-agnostic default, that works for me.

milktrader:
I stand corrected. Financial time series will have the last timestamp and the last value. So daily data collapsed to weekly would be the Friday timestamp and the last value recorded for the week (which is Friday's value). 

As long as the defaults are documented, I don't see a major issue here with selecting defaults as long as there are use cases to support them. Maybe to add some extra documentation coverage we can add in REPL documentation for this one?

Eventually, we should have in REPL documentation for all the methods of course.

GordStephen:
Thinking about this some more, what if the timestamp aggregator was given first and the value aggregator defaulted to the same function?

```julia
collapse{T,N,D}(ta::TimeArray{T,N,D}, period::Function=week, timestamp::Function=last, value::Function=timestamp)
```

That would simplify calling things like (first timestamp, first value) operations (presumably fairly common) without affecting calls like (first timestamp, mean value).

```julia
collapse(ta, hour, first)
collapse(ta, hour, first, mean)
```

Whatever we go with, this will likely involve a change in method signature that will be hard to automatically detect without parametrized function types  - I think that's the main barrier to overcome before this can be merged. A simple solution might be to just assign the new method to a whole new function name (maybe just temporarily) and provide a deprecation warning on the existing collapse method (which would just wrap the new function for the time being)?

GordStephen:
Just added the aforementioned type signature change:

```julia
collapse(ta::TimeArray, period::Function=week, timestamp::Function=last, value::Function=timestamp)
```

I also changed the tests around for the new argument order. Last thing to do before this can be pulled is figure out how to smoothly deprecate the old arguments...

milktrader:
Yep, and this will prompt a new minor version because of these changes.

I'd say to add the old method to the deprecated.jl file, and tag the final `0.7.x` version and then we can pull the trigger with `0.8.0`.

GordStephen:
So I finally have some time to get back to this (better late than never!) - the tricky part with the deprecation is that the new method doesn't have a distinct type signature from the old method, so we can't support both simultaneously. If we move the current method to deprecated.jl and push `0.7.5` with a warning, people won't be able to switch over to the new API until `0.8.0` anyways, at which point their code will already be breaking...

Some options I see:

1. Drop the defaults for `period` and `timestamp` in the new method, resulting in a unique type signature, and keep but deprecate the old method
2. Change the function name of the new method to something other than `collapse` and deprecate `collapse` (we could eventually switch back somewhere down the road, although that would require deprecating the new function name as well)
3. Just remove the old method and break existing code

Personally I like the first option - it's easy for end-users to add back in personalized defaults through a wrapper function anyway:

```julia
mycollapse(ta::TimeArray, period::Function=mydefaultperiod, timestamp::Function=mydefaulttimestamp, value::Function=mydefaultvalue) = collapse(ta, period, timestamp, value)
```

...or better yet, a wrapper package that defines useful defaults for a relevant application (e.g. MyUseCaseTimeSeries.jl as a thin wrapper that re-exports TimeSeries.jl)

milktrader:
How about use the rc file?

milktrader:
Also, for breaking releases, we'll just up the number to `0.8.0` and give legacy users the option to enforce only earlier versions in their code.

GordStephen:
Good point about using the rc file for setting defaults. Would you support Option 1 then?

milktrader:
ok, to compare things, here is the two versions we're talking about ...

````julia
# new
function collapse{T,N,D}(ta::TimeArray{T,N,D}, period::Function=week, timestamp::Function=last, value::Function=timestamp)

# current
function collapse{T,N,D}(ta::TimeArray{T,N,D}, f::Function; period::Function=week)
````

milktrader:
Yeah, option 1 looks solid. This means the new type signature would be this?

````julia
function collapse{T,N,D}(ta::TimeArray{T,N,D}, period::Function, timestamp::Function, value::Function=timestamp)
````

GordStephen:
Yup, exactly. I'll make the final changes and push them here.

milktrader:
lgtm, pull the trigger and let there be 0.8.0!

GordStephen:
METADATA.jl PR - https://github.com/JuliaLang/METADATA.jl/pull/5098

Issue #250, opened 2016-02-27
Liberate TimeArrays from TimeTypes/Vectors


gajomi:
I would like to suggest that it is an "issue" that timestamps are constrained to be `Vector{D<:TimeType}`. Be forewarned that there is a wall of text that follows [ hopefully I am not arguing too forcefully :) ].

Arguments for relaxing `TimeType` eltype restriction:

1. Not all recorded times will be `TimeTypes`. Most every laboratory I have been in that develops their own equipment usually ends up rolling out one or more more custom notion of time. These may not always be easy to map onto an extant concrete `TimeType` (as Stefan Karpinski recently brought up [here]( https://groups.google.com/forum/#!searchin/julia-users/time|sort:date/julia-users/BT4bP5VyaRY/Ax7IzLc_BAAJ), "time" could depend on [esoteric details of the earth's rotation](https://en.wikipedia.org/wiki/%CE%94T)). It is not uncommon in these cases to represent time by some numeric primitive or a simple data structure, and it would be nice if `TimeArray`s easily accommodated this. Having said that the various `TimeType` types could still be the official supported timestamps in the package.
2. In a simulation context there is oftentimes no natural starting epoch from which to count so it is natural to start from one or zero. Furthermore it is common to non-dimensionalize the model equations so that the simulation "time" parameter is dimensionless.
3. Sometimes you want floating point timestamps. Floating point timestamps are quite natural for many models (e.g. point processes on the real line, intervals of an adaptive ode integrator). Floating poitn types can also make sense in experimental contexts when one is merging together timestamps from data at wildly different scales (as in geophysics, cosmology, etc.).

Argument for relaxing `Vector` container restriction:

The main reason is that it would allow one to encode information about the sampling in the timestamp type. A key example would be allowing for `Range`-like containers:
```julia
julia> using Base.Dates
julia> TimeArray(range(now(),Second(1),5),0:4,[""])
5x1 TimeSeries.TimeArray{Int64,1,DateTime,StepRange{DateTime,Base.Dates.Second},UnitRange{Int64}} 2016-02-26T17:57:55 to 2016-02-26T17:57:59
2016-02-26T17:57:55 | 0       
2016-02-26T17:57:56 | 1       
2016-02-26T17:57:57 | 2       
2016-02-26T17:57:58 | 3       
2016-02-26T17:57:59 | 4 
```
[...]

PR #251, opened 2016-02-27
WIP: Removed Vector{D<:TimeType} constraint on timestamp construction

gajomi:
This is a proof of concept for a set of features to solve #250. The main work so far is to remove the constrain that timestamps be `Vector{D<:TimeType}`. Examples of new functionality can be found in the test. On this note, I have to confess that this is my first time seeing Midje-like testing, so I am not so sure about the logical organization of these tests. Any suggestions would be appreciated

On a more substantial note, the current implementation of this behavior adds one more type parameter to `TimeArray` to encode the timestamp container type. So it start to look like a bunch of parameters. However, in my first pass through the source, I didn't see anywhere were it would be strictly needed to have the time eltype available to a method, as this could be read from the timestamp container. So it might be possible to remove the D parameter from the type.

### DIFF
--- a/src/timearray.jl
+++ b/src/timearray.jl
@@ -4,44 +4,44 @@ import Base: convert, length, show, getindex, start, next, done, isempty, endof
 
 abstract AbstractTimeSeries
 
-immutable TimeArray{T, N, D<:TimeType, A<:AbstractArray} <: AbstractTimeSeries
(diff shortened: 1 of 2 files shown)


milktrader:
@gajomi for experimental branches, it would be a good idea to push this to this repo as a branch. This way, others can experiment with it and the like without it disrupting the main branch.

Issue #249, opened 2016-02-22
LoadError: metadata doesn't match


tcfuji:
I downloaded the most recent version and tried the examples in the docs. In the `merge` section, I'm getting the following:
```julia
Caterapple = merge(AAPL,CAT);

LoadError: metadata doesn't match
while loading In[73], in expression starting on line 1

 in merge at /Users/tcf/.julia/v0.4/TimeSeries/src/combine.jl:8 (repeats 2 times)
```
Is this behavior intended?

GordStephen:
Yes, although it's probably more of a bug than a feature (https://github.com/JuliaStats/TimeSeries.jl/issues/164) - the docs should be fixed though, thanks for pointing that out. There also doesn't seem to be a way to specify new metdata during a merge anymore (e.g. ` merge(AAPL, CAT, meta="Caterapple")`, which is most definitely a bug that needs fixing.

milktrader:
I've always thought to implement a `Dict` in this case.

````julia
julia> foo = Dict(AAPL=>"AAPL", CAT=>"CAT")
Dict{TimeSeries.TimeArray{Float64,2,Date,Array{Float64,2}},ASCIIString} with 2 entries:
  8336x12 TimeSeries.Time… => "AAPL"
  13090x12 TimeSeries.Tim… => "CAT"
````
The rationale being that there is some trail of what data is in the current TimeArray. It doesn't solve which columns are from which TimeArrays but does provide something meaningful if incomplete.

**Comment by milktrader** [comment]
Maybe a quick solution, since this is breaking, is to have a default assignment of the meta field as an empty string `""` and allow users to specify it as something different if they choose. They will lose the meta data from the corresponding TimeArrays of course.

**Comment by GordStephen** [comment]
Just submitted #253 to address this via default assignment while also adding the option for explicit meta redefinition. I also extended that default behaviour to math operations for consistency, although the custom meta option obviously won't work in that case...

**Comment by GordStephen** [comment]
Closing this as I just pulled #253 which fixes the original issue here - discussion about different behaviour possibilities can continue in #164.

```julia
julia> Caterapple = merge(AAPL,CAT);

julia> Caterapple.meta
Void
```

milktrader:
good default and we can improve from here. +1

PR #253, opened 2016-02-27
Allow operations on TimeArrays with non-matching meta - address #249

GordStephen:
Added default (`Void`) metadata to merge and math operation results when input metas don't match. Also allows for meta to be set explicitly during merge.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 * allow math operations between different Number subtypes
 * explicitly convert column names to strings  in `readtimearray`
+* operations between TimeArrays with non-matching meta fields now succeed, with a `Void` meta in the result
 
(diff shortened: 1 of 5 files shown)


milktrader:
Looks good. Deserves a mention in NEWS.md and a tagged version

milktrader:
Nice work @GordStephen ! This should just be tagged and sent to METADATA. 

Since you have the latest commit, you can do the following ...

````julia
julia> Pkg.tag("TimeSeries")

julia> Pkg.update()

julia> Pkg.publish()

# then push some buttons on the github API and @tkelman should merge it soon.
````

GordStephen:
Cool, done - https://github.com/JuliaLang/METADATA.jl/pull/4749

PR #254, opened 2016-02-27
Explicitly convert column headings to strings in readtimearray

GordStephen:
Fixes an issue where column headings might not be automatically interpreted as strings during file loading (for example, if column names are numbers).


milktrader:
I'd say update NEWS with 0.7.4 (unreleased) and merge.

milktrader:
sweet

PR #258, opened 2016-05-04
Don't do using Dates outside the top package module

tkelman:
This pulls in the exports outside of the package whenever anyone uses it, rather than just inside the package's module.


milktrader:
Tagged 0.8.1 with these changes. Should I do a PR with METADATA now?

tkelman:
this is a cleanliness thing, so no big urgency

milktrader:
Ok, I'll just do it now then since it's ready

PR #261, opened 2016-05-10
More robust to() and from()

dourouc05:
Useful when the time series is empty (e.g. due to a previous filtering with `to` or `from`).

### DIFF
--- a/test/split.jl
+++ b/test/split.jl
@@ -17,9 +17,14 @@ end
 
 facts("split date operations") do
 
-    context("from and to correctly subset") do
+    context("from and to correctly subset") do # Including zero-length time series
         @fact length(from(cl, Date(2001,12,28))) --> 2
+        @fact length(from(cl, Date(2002,1,1)))   --> 0
+        @fact length(from(from(cl, Date(2002,1,1)), Date(2012,1,1))) --> 0 
+        
         @fact length(to(cl, Date(2000,1,4)))     --> 2
+        @fact length(to(cl, Date(1999,1,4)))     --> 0 
+        @fact length(to(to(cl, Date(1999,1,4)), Date(1912,1,1))) --> 0 
     end
         
     context("when method correctly subset") do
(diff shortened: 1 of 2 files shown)


milktrader:
Should this not throw an error instead?

dourouc05:
I don't think it makes sense to throw an error in this case. Actually, here is my use case: 
`to(from(ts, beginDate), endDate)`
where ts has dates that are sometimes all before beginDate. If `to()` and `from()` start throwing errors, I'll have to add some nontrivial error checking code on my side… (something like `min(timestamp(ts)) >= beginDate`, which does not )

Also, throwing an error is inconsistent with existing Julia behaviour, e.g. filtering an empty vector does not throw any error, but returns an empty vector: 
`filter(x -> x > 0, []) == []`

milktrader:
Ok, good points. I'm on board. Can you add two tests and I'll merge it.

Somewhere in the `to` and `from` blocks, something like ...

````julia
context("from and to correctly subset") do
        
     @fact length(from(cl, Date(2001,12,28))) --> 2
@fact length(from(cl, Date(2002,1,1))) --> 0

        @fact length(to(cl, Date(2000,1,4)))     --> 2
@fact length(to(cl, Date(1999,1,4)))     --> 0

    end
````

**Comment by dourouc05** [comment]
Here it is. I mainly added tests when the zero-length arrays gets into to and from.

File: test/split.jl [hunk 1]
```diff
@@ -17,9 +17,14 @@ end
 
 facts("split date operations") do
 
-    context("from and to correctly subset") do
+    context("from and to correctly subset") do # Including zero-length time series
```
milktrader: instead of commenting on the test, it would be better to add that into the `context` string so it becomes transparent. I can change this though, thanks for the tests!

milktrader:
Alright then, I'll merge since there hasn't been any objection for a few days except my objection, which has been overcome. Thanks!

dourouc05:
By the way, could you publish a new version of the package with this fix, please?

milktrader:
Yep, I see it's pending. Will do it.

dourouc05:
Great, thank you!

PR #270, opened 2016-07-30
Use dirname(@__FILE__) instead of Pkg.dir

tkelman:
This allows installing the package elsewhere.

### DIFF
--- a/test/readwrite.jl
+++ b/test/readwrite.jl
@@ -13,11 +13,11 @@ facts("readwrite parses csv file correctly") do
     end
 
     context("Specifying DateTime string format for reading") do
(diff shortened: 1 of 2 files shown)


milktrader:
Sweet! Thanks @tkelman! Looks like this only affects tests so I'll not tag this.

tkelman:
If there were a rush I'd let you know. Note that there's an 0.5 rc that you can now test against, I didn't add that to your .travis.yml file here because you have an explicit upper julia version bound.

milktrader:
Yep, the bound is in place because I wasn't sure I could commit time to keep pace with fast-moving changes, but if we're rc now then I guess the time has arrived to start testing.

PR #272, opened 2016-08-02
add customizable representation of missing values

milktrader:


milktrader:
I'll merge this since it doesn't break anything, new tests and docs are provided, and there are no surprises to earlier tagged versions

Issue #267, opened 2016-07-19
Rename columns


femtotrader:
After doing

```julia
start = Date(2000,1,30)
periods = 9
freq = Dates.Day(1)
dates = start:start + (periods - 1) * freq
mytime3 = TimeArray(collect(dates), rand(length(dates), 3), ["col1", "col2", "col3"])
```

I was looking for a way to rename columns.
I thought that some methods such as `names!` and `rename!` (like [DataFrames.jl](https://dataframesjl.readthedocs.org/)) would support `TimeArray` but it doesn't seems to be the case.

```julia
julia> names!(mytime3, ["c1", "c2", "c3"])
ERROR: MethodError: `names!` has no method matching names!(::TimeSeries.TimeArray{Float64,2,Date,Array{Float64,2}}, ::Array{ASCIIString,1})
Closest candidates are:
  names!(::DataFrames.AbstractDataFrame, ::Any)
```

and

```julia
julia> rename!(mytime3, Dict("col1"=>"c1"))
ERROR: MethodError: `rename!` has no method matching rename!(::TimeSeries.TimeArray{Float64,2,Date,Array{Float64,2}}, ::Dict{ASCIIString,ASCIIString})
Closest candidates are:
  rename!(::DataFrames.Index, ::Any)
  rename!(::DataFrames.Index, ::Any, ::Any)
  rename!(::DataFrames.AbstractDataFrame, ::Any...)
```

milktrader:
The `TimeArray` is `immutable` so you can't change things. You can create a new TImeArray though. This immutability feature is by design.

dourouc05:
I think there is still something to do here: how do you **easily** create a copy of a time series by changing only a few fields? The only syntax I can think of is: 
```jl
TimeArray(timestamp(ta), values(ta), ["col1", "col2"], meta(ta))
```
In Scala, a language that promotes immutability as a core concept, you would use a copy() function (automatically defined for case classes: http://www.scala-lang.org/docu/files/ScalaReference.pdf, section 5.3.2, page 68); adapted to Julia, this would give: 
```jl
copy(ta, colnames=["col1", "col2"])
```
I think it's a much clearer and cleaner way that creating a new TimeArray from scratch using the constructor. (What if there is a new field for TimeArrays?)

The implementation could be quite simple: 
```jl
copy(ta; timestamp=ta.timestamp, values=ta.values, colnames=ta.colnames, meta=ta.meta) = 
[...]

PR #275, opened 2016-08-09
Update, rename, head, tail methods added

milktrader:


milktrader:
https://github.com/JuliaStats/TimeSeries.jl/issues/267

Issue #283, opened 2016-09-13
Enable update to work on empty TimeArrays


ElevatorDeLuxe:
Hello, everybody!

I have the following issue:
If you create an empty TimeArray and then call an update, you receive an error.

Example:
`time_array = TimeArray(Vector{Date}(), Array{Float64}(0))`
`update(time_array, Date(2000,1,1), 42.0)`
results in the error
`ERROR: ArgumentError: reducing over an empty collection is not allowed
 in _mapreduce at reduce.jl:139
 in update at /home/user/.julia/v0.4/TimeSeries/src/modify.jl:14`

The option to update an empty TimeArray would be quite handy. Maybe I find the time to fix it myself. I would then commit a pull request.

Best regards Thomas

milktrader:
Why would you rather update an empty TimeArray versus creating it?

If you update an empty TimeArray, you first need to create it and then update it versus just creating the TimeArray with values from the update to begin with.

milktrader:
The code to support this would be simple ...
````julia
elseif length(ta) == 0
    TimeArray(tstamp, val)
else 
...
````

ElevatorDeLuxe:
Okay. I think this depends on the definition of creation. I would say 
` time_array = TimeArray(Vector{Date}(), Array{Float64}(0)) `
is already creating it even it it is empty. For the workflow of a function that returns a TimeArray, that can potentially be empty, this would be a nice feature.

`function returnTimeArray(someParams...)`
`   time_array = TimeArray(Vector{Date}(), Array{Float64}(0))`
`   some_logic_that_adds_entries(time_array, someParams...)` # possibly no entries are added
`   return time_array`
`end`

ElevatorDeLuxe:
Commited pull request #285 .

milktrader:
This was closed as won't fix.

PR #285, opened 2016-09-14
Enable update to work on empty TimeArrays #283

ElevatorDeLuxe:


src/modify.jl
```diff
@@ -1,7 +1,10 @@
 ###### update ####################
 
 function update{T,N,D}(ta::TimeArray{T,N,D}, tstamp::D, val::Vector{T})
-    if tstamp < maximum(ta.timestamp)
+    if length(ta) == 0
+        t    = vcat(tstamp)
```
ararslan: Why vcat? I think vec would be more appropriate, assuming that's what you're trying to do.

ararslan:
Should add a test for this functionality

src/modify.jl
milktrader: There are constructors already in place to take care of a single `tstamp`. So you really just want to construct the TimeArray with a simple `TimeArray(tstamp, val)` and you're done. 

I'm not sure if you noticed by I assigned this issue to myself, just haven't gotten to it yet. And it will require a few more tests as well.

milktrader:
I pulled this PR and then pushed the changes but it didn't update this PR, so I'll close this in favor of #286

PR #286, opened 2016-09-14
Allow update on empty TimeArray

milktrader:
Originally pull request #285

### DIFF
--- a/src/modify.jl
+++ b/src/modify.jl
@@ -1,29 +1,31 @@
 ###### update ####################
 
-function update{T,N,D}(ta::TimeArray{T,N,D}, tstamp::D, val::Vector{T})
-    if length(ta) == 0
-        t    = vcat(tstamp)
-        vals = vcat(val')
-    elseif tstamp < maximum(ta.timestamp)
-        error("only appending operations supported")
-    else
-        t    = vcat(ta.timestamp, tstamp)
-        vals = vcat(ta.values, val')
-    end
-    TimeArray(t, vals, ta.colnames, ta.meta)
-end
+function update{T,N,D}(ta::TimeArray{T,N,D}, tstamp::D, val::Array{T,N})
 
-function update{T,N,D}(ta::TimeArray{T,N,D}, tstamp::D, val::T)
     if length(ta) == 0
-        t    = vcat(tstamp)
-        vals = vcat(val)
+        uta = TimeArray(tstamp, val, ta.colnames, ta.meta)
     elseif tstamp < maximum(ta.timestamp)
         error("only appending operations supported")
     else
         t    = vcat(ta.timestamp, tstamp)
         vals = vcat(ta.values, val)
+        uta  = TimeArray(t, vals, ta.colnames, ta.meta)
     end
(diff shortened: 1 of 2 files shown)


milktrader:
On the plus side for this PR, the `update` method was cleaned up and the need for array inverting was removed. 

On the not-so-plus side, this feature leaves it up to the user to define an empty array, whose syntax is not obvious.

Maybe create empty array constructors too?

ElevatorDeLuxe:
An empty array constructor would be consistent with the new update function. In my opinion, it would be useful. 

By the way, thanks for the work.

milktrader:
I'm a little uneasy about this to be honest. A TimeArray structure is an immutable object and it seems strange to create an empty one if you start with the idea that it's immutable. Empty TimeArrays were enabled to return the result of a filtering method. 

`update` is a strange concept to begin with because it doesn't really __update__ a data structure object, but rather creates a separate one. In practical use, time series are naturally added to as time passes so there is a use for it.

I need to think about this problem a little more and still wonder why it's important to initialize an empty TimeArray.

ararslan:
> A TimeArray structure is an immutable object

What's the reasoning behind that? I figured that `AbstractArray`s were in general expected to be mutable, but I could be way off on that. It seems that by making it immutable you lose out on some of the performance benefits of in-place operations that don't require making copies.

I don't see the harm in being able to initialize an empty `TimeArray`, but I also don't see the benefit if `TimeArray`s are immutable. That isn't to say I don't think it should be possible though.

milktrader:
Once you create a data structure like a TimeArray and it is not immutable, you are free to engage in all sorts of chicanery, most of the time without you knowing it. It's a safeguard against chaos.

````julia
julia> using MarketData

julia> cl.values = reverse(cl.values)
ERROR: type TimeArray is immutable

julia> cl.values = vcat(cl.values, 12.12)
ERROR: type TimeArray is immutable

julia> cl.colnames = ["Open"]
ERROR: type TimeArray is immutable
````
Yes, the field values themselves are mutable, but the type is not. There was evidence that an immutable type had performance gains over a mutable type when it was first introduced.

There is certainly room for a separate time-based type where one could enforce invariants and append to existing structure, but that is not what a TimeArray is. You might design such a different time-based structure and call it a `TimeStream` but I'm not entirely clear what that would look like.

milktrader:
The holy grail for a time series structure is to not have it be a type, but rather to have it be __just__ an array. For this to work, the `Integer` that defines the row of an array would need to be replaced with a `TimeType`.

milktrader:
>For example:
>function returnTimeArray(someParams...)
t>ime_array = TimeArray(Vector{Date}(), Array{Float64}(0))
>some_logic_that_adds_entries(time_array, someParams...) # possibly no entries are added
>return time_array
>end

For reference, this is the use-case to support updating empty time arrays, provided by @ElevatorDeLuxe

milktrader:
This discussion isn't closed with the push of commits to make updating an empty array an error. The previous code to allow this behavior is commented out.

milktrader:
Well no objection to this won't fix PR after 9 days so merging it.

PR #298, opened 2016-11-09
ENH: readtimearray can accept several kind of input 

### LINKED ISSUE #297: readtimearray should accept also IOBuffer as first parameter
readtimearray currently accept fname as AbstractString

```julia
    function readtimearray(fname::AbstractString; delim::Char=',', meta=nothing, format::AbstractString="")
        cfile = readdlm(fname, delim)
        ...
    end
```

this is a bit too restrictive

readtimearray should accept also IOBuffer as first parameter

```julia

    function readtimearray(source; delim::Char=',', meta=nothing, format::AbstractString="")
        cfile = readdlm(source, delim)
        ...
    end
```

So it will be possible to provide either a filename or a buffer (like `readdlm`)

femtotrader:
Closes #297 

IOBuffer for example can be an allowed input

### DIFF
--- a/test/readwrite.jl
+++ b/test/readwrite.jl
@@ -1,6 +1,8 @@
-using TimeSeries, MarketData
+using FactCheck
 FactCheck.setstyle(:compact)
 FactCheck.onlystats(true)
+
+using TimeSeries, MarketData
 
 facts("readwrite parses csv file correctly") do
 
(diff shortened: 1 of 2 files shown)


test/readwrite.jl
```diff
@@ -1,7 +1,9 @@
-using TimeSeries, MarketData
+using FactCheck
```
milktrader: In the future, please don't make changes in one test file that isn't reflected in a different test files and submit that as PR. It introduces inconsistencies. In this case, I'll merge the PR and remedy it.

Issue #302, opened 2017-02-04
Support a time series recipe to work with StatsPlots.jl


milktrader:
https://github.com/JuliaPlots/StatPlots.jl/issues/37#issuecomment-277462579

milktrader:
- [ ] write the code for the recipe
- [ ] document its use
- [ ] write some unit tests
- [ ] add new package requirements to REQUIRE
- [ ] update NEWS
- [ ] tag as a new minor version

mkborregaard:
https://github.com/JuliaStats/TimeSeries.jl/pull/303

mkborregaard:
points 1 and 4 above are covered in #303. With regards to 3, unit tests, recipe testing requires Plots (which there is otherwise no reason to depend on).

PR #303, opened 2017-02-09
added basic plot recipe

mkborregaard:
This adds basic plotting functionality to TimeArray objects, with a very light dependency (RecipesBase a single macro). It allows plotting as simply as `plot(ohlc)`.
Discussed here https://github.com/JuliaPlots/StatPlots.jl/issues/37#issuecomment-278708392 and here https://github.com/JuliaStats/TimeSeries.jl/issues/302 .

The current recipe will plot all values in the array, which may not make sense if they are not on the same scale, but you can always `plot(ohlcv["Volume"])`.

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -1,6 +1,6 @@
 
 @recipe function f{T<:TimeArray}(ta::T)
-       labels --> reshape(ta.colnames,1,length(ta.colnames))
-       seriestype --> :path
-       ta.timestamp, ta.values
+    labels --> reshape(ta.colnames,1,length(ta.colnames))
+    seriestype --> :path
+    ta.timestamp, ta.values
 end
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -1,0 +1,6 @@
+
+@recipe function f{T<:TimeArray}(ta::T)
+       labels --> reshape(ta.colnames,1,length(ta.colnames))
+       seriestype --> :path
+       ta.timestamp, ta.values
+end
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,1 +1,2 @@
 julia 0.5
+RecipesBase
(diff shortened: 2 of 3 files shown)


src/plotrecipes.jl
```diff
@@ -0,0 +1,6 @@
+
+@recipe function f{T<:TimeArray}(ta::T)
+       labels --> reshape(ta.colnames,1,length(ta.colnames))
+       seriestype --> :path
+       ta.timestamp, ta.values
```
ararslan: Please use 4 space indents for consistency with the rest of the repository

milktrader:
Coverage will be fine maybe increase with some tests

mkborregaard:
Do you need me to write a line for the docs (in which case where?) or will you leave this implicit?

milktrader:
If you can think of a reasonable test it would be great. But if it doesn't do anything but move the coverage then no.

dourouc05:
@mkborregaard: As a user, I'd appreciate a bit of doc about this (at least, say that it exists, maybe briefly how to make a basic plot — i.e. `using Plots` and `plot(time series)` — and a link to Plots).

mkborregaard:
Sure - where in the docs should it go?

dourouc05:
In [getting started](http://timeseriesjl.readthedocs.io/en/latest/getting_started.html)? [Customize how TimeArray is displayed](http://timeseriesjl.readthedocs.io/en/latest/dotfile.html) could also be interesting, but I don't feel at ease with the idea of putting plots there.

milktrader:
I was planning a new section called simply __Plotting__

milktrader:
If you want to start a file called `plotting.rst` I can make the links. I would copy an existing `.rst` file for the format template.

mkborregaard:
No probs. It will be a little while cause I'm going on holiday now.

mkborregaard:
Here you are https://github.com/JuliaStats/TimeSeries.jl/pull/305 
I have started it up, can you take it from here?

PR #305, opened 2017-02-21
Plotting documentation

mkborregaard:
As requested here https://github.com/JuliaStats/TimeSeries.jl/pull/303#issuecomment-279829409 I have added some documentation for using the Plots recipe.
This is my first experience with restructured text so it may need some editing.


milktrader:
Good start

PR #306, opened 2017-02-24
Candlesticks

mkborregaard:
I implemented support for candlestick and Heikin-Ashi candlestick plots for OHLC data using the links given by @milktrader here: https://github.com/JuliaPlots/StatPlots.jl/issues/37
OHLC appears to constitute quite a lot of the data currently associated with TimeSeries.
The recipe looks in the colnames for “open” “low” “high” and “close” (ignoring case) and throws an error if it does not find them. There are possibly better ways of doing this but this code (the extract_ohlc function) is straightforward to adapt.

I thought this would take me an hour but turned out I was really wrong :-)
I think it is best if the documentation is adapted by one of you who work with this data type. It should be straightforward from the examples given below.

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -14,129 +14,48 @@
         bw == nothing && (bw = 0.8)
         bar_width := bw / 2 * minimum(diff(unique(Int.(cs.time))))
 
-        len = length(cs.open)
-        series = Vector{Int}(len)
-        series[1] = cs.close[1] >= cs.open[1] ? 3 : 1
+        # allow passing alternative colors as a vector
+        cols = get(d, :seriescolor, nothing)
+        cols = (isa(cols, Vector{Symbol}) && length(cols) == 2) ? cols : [:red, :blue]
 
-        # create four series
-        for i in 2:len
-            fil = 2*(cs.close[i] >= cs.open[i])
-            col = 1 + (cs.close[i] >= cs.close[i-1])
-            series[i] = col + fil
-        end
+        attributes = [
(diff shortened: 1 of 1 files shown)


mkborregaard:
To make a candlestick plot, simply specify `seriestype` (or `st` or `typ`, whatever you prefer) as `:candlestick` when plotting a time array with OHLC columns in it.
```julia
using MarketData, TimeSeries, Plots
plot(MarketData.ohlc[130:180], seriestype = :candlestick)
```
skaermbillede 2017-02-24 kl 23 20 43

To get Heikin-Ashi candlesticks, instead specify `:heikinashi`:
```julia
plot(MarketData.ohlc[130:180], seriestype = :heikinashi, color = [:red :black])
```
skaermbillede 2017-02-24 kl 23 33 59

src/plotrecipes.jl
```diff
@@ -1,6 +1,98 @@
...
+type Candlestick{D <: TimeType}
+    time::Vector{D}
+    open::AbstractVector
+    high::AbstractVector
+    low::AbstractVector
+    close::AbstractVector
+end
+
+Candlestick(ta::TimeArray) = Candlestick(extract_ohlc(ta)...)
+
```
mkborregaard: These two functions could be written tests for. It will be awhile before I find time to look at this again, but I don't feel strongly it has to be me who writes them.

milktrader:
This is great stuff, thanks!

Czechking:
Awesome stuff....thanks!

PR #311, opened 2017-05-02
 timearray: add StepRange{Int,Int} support for getindex

iblis17:
```julia
    julia> cl[1:2:10]
    5x1 TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}} 2000-01-03 to 2000-01-13

                 Close
    2000-01-03 | 111.94
    2000-01-05 | 104.0
    2000-01-07 | 99.5
    2000-01-11 | 92.75
    2000-01-13 | 96.75
```


src/timearray.jl
```diff
@@ -199,16 +199,20 @@ function getindex{T,N,D}(ta::TimeArray{T,N,D}, d::D)
...
     dates = sort(dates)
     idxs, _ = overlaps(ta.timestamp, dates)
     ta[idxs]
 end #getindex
 
+# StepRange{Date,...}
 getindex{T,N,D}(ta::TimeArray{T,N,D}, r::StepRange{D}) = ta[collect(r)]
 
+# StepRange{Int,Int}
+getindex{T,N}(ta::TimeArray{T,N}, r::StepRange{Int,Int}) = ta[collect(r)]
```
ararslan: Any `Integer` subtype and any integer-valued array can be used for indexing; it needn't be limited to `StepRange`s. In this case we probably want to limit indexing to one-dimensional collections. So this could be written more generally as
```julia
getindex{T,N,R<:Integer}(ta::TimeArray{T,N}, r::AbstractVector{R}) = ta[r]
```
Also note that it doesn't need to be `collect`ed for indexing.

ararslan: However, a better solution may be to refactor the existing methods to widen the signatures to `AbstractVector` for `Integer` subtypes.

iblis17: I change the signatures. 
Committed @ 24b3410

src/timearray.jl
```diff
@@ -173,16 +173,16 @@ function getindex{T,D}(ta::TimeArray{T,1,D}, r::UnitRange{Int})
 end
 
 # array of rows
-function getindex{T,N,D}(ta::TimeArray{T,N,D}, a::Vector{Int})
+function getindex{T,N,D}(ta::TimeArray{T,N,D}, a::AbstractVector{Int})
```
ararslan: We should widen this to any integer subtype:
```julia
function getindex{T,N,D,S<:Integer}(ta::TimeArray{T,N,D}, a::AbstractVector{S})
```
here and below.

ararslan:
It would be great if you could add tests for the new behavior.

src/timearray.jl
iblis17: 👌  commit @ 2bdeeed

test/timearray.jl
```diff
@@ -116,38 +116,40 @@ facts("ordered collection methods") do
...
     context("getindex on array of Int and Date") do
         @fact ohlc[[1,10]].timestamp                           --> [Date(2000,1,3), Date(2000,1,14)]
         @fact ohlc[[Date(2000,1,3),Date(2000,1,14)]].timestamp --> [Date(2000,1,3), Date(2000,1,14)]
     end
-  
+
     context("getindex on range of Int and Date") do
         @fact ohlc[1:2].timestamp                                  --> [Date(2000,1,3), Date(2000,1,4)]
+        @fact ohlc[1:2:4].timestamp                                --> [Date(2000,1,3), Date(2000,1,5)]
+        @fact ohlc[Int8(1):Int8(2):Int8(4)].timestamp              --> [Date(2000,1,3), Date(2000,1,5)]
```
iblis17: simple test cases is here.

milktrader:
Thanks!

milktrader:
We need someone to volunteer to push new version and update the NEWS. I will put it on my list but no promises for a few weeks.

iblis17:
I added this issue to NEWS.

ararslan:
I can make a new release.

PR #312, opened 2017-05-09
doc: fix rst code block syntax

iblis17:


docs/source/plotting.rst
```diff
@@ -10,13 +10,15 @@ plot
 The recipe allows TimeArray objects to be passed as input to ``plot``. The recipe will plot each variable as an individual
 line, aligning all variables to the same y axis (here shown using PlotlyJS as a plotting backend).
 
+::
+
     using Plots, MarketData, TimeSeries
     plotlyjs()
-    plot(MarketData.ohlc)
```
ararslan: Is `ohlc` exported from the MarketData package?

iblis17: Yes, it is.
https://github.com/JuliaQuant/MarketData.jl/blob/master/src/MarketData.jl#L8

PR #313, opened 2017-06-01
allow views of timestamps in constructors

JeffreySarnoff:
allows TimeArrays to be constructed from view_s of another TimeArray's timestamps
just as TimeArrays can be constructed from view_s of another TimeArray's values

### DIFF
--- a/test/timearray.jl
+++ b/test/timearray.jl
@@ -56,16 +56,18 @@ facts("type constructors enforce invariants") do
 end
 
 facts("type constructors allow views") do
-    
+
     source_rows = 101:121
-    source_cols = 1:size(AAPL.values)[2]    
-    
-    AAPL1 = TimeArray(AAPL.timestamp[source_rows], AAPL.values[source_rows, source_cols], AAPL.colnames, AAPL.meta)
-    
+    source_cols = 1:size(AAPL.values)[2]
+
+    AAPL1 = TimeArray(AAPL.timestamp[source_rows],
+                      AAPL.values[source_rows, source_cols],
+                      AAPL.colnames, AAPL.meta)
+
     tstamps = view(AAPL.timestamp, source_rows)
     tvalues = view(AAPL.values, source_rows, source_cols)
-    
-    APPL2 = TimeArray(tstamps, tvalues, AAPL.colnames, AAPL.meta)
+
+    AAPL2 = TimeArray(tstamps, tvalues, AAPL.colnames, AAPL.meta)
 
     @fact length(AAPL1) --> length(AAPL2)
     @fact AAPL1[5].values[3] --> AAPL2[5].values[3]
@@ -76,8 +78,8 @@ facts("type constructors allow views") do
     context("match final values") do
         @fact AAPL[121].values --> AAPL2[end].values)
     end
-=#   
-end    
+=#
+end
 
 facts("construction without colnames") do
 
--- a/test/timearray.jl
+++ b/test/timearray.jl
@@ -57,21 +57,23 @@ end
 
 facts("type constructors allow views") do
     
-    items = 101:121
-    ncols = 1:size(AAPL.values)[2]
-    tstamps = view(AAPL.timestamp, items)
-    tvalues = view(AAPL.values, items, ncols)
+    source_rows = 101:121
+    source_cols = 1:size(AAPL.values)[2]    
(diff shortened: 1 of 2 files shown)


ararslan:
Nice, thanks! Could you add a test for the new functionality?

JeffreySarnoff:
I have not used that style of test --  the old way the test would be e.g.

```julia
    items = 101:121
    cols = 1:size(AAPL.values)[2]
    tstamps = view(AAPL.timestamp, items)
    tvalues = view(AAPL.values, items, cols)
    APPL2 = TimeArray(tstamps, tvalues, AAPL.colnames, AAPL.meta)
    
    @test AAPL[101:121] == AAPL2[1:end]
```

test/timearray.jl
```diff
@@ -55,6 +55,16 @@ facts("type constructors enforce invariants") do
...
 end
 
+facts("type constructors allow views") do    
+    items = 101:121
+    cols = 1:size(AAPL.values)[2]
+    tstamps = view(AAPL.timestamp, items)
+    tvalues = view(AAPL.values, items, cols)
+    APPL2 = TimeArray(tstamps, tvalues, AAPL.colnames, AAPL.meta)
+    
+    @test AAPL[101:121] == AAPL2[1:end]
```
iblis17: there is a macro `@fact` from FactCheck...

ararslan:
Yes, as @iblis17 points out, this repository unfortunately still uses the outdated FactCheck package for its testing rather than the unit testing functionality provided in Base. To turn a `@test` into a `@fact`, you do `@fact a --> b` rather than `@test a == b`.

JeffreySarnoff:
I do not see the problem -- it works for me with cut and paste

test/timearray.jl
```diff
@@ -55,6 +55,30 @@ facts("type constructors enforce invariants") do
...
+    
+    source_rows = 101:121
+    source_cols = 1:size(AAPL.values)[2]    
+    
+    AAPL1 = TimeArray(AAPL.timestamp[source_rows], AAPL.values[source_rows, source_cols], AAPL.colnames, AAPL.meta)
+    
+    tstamps = view(AAPL.timestamp, source_rows)
+    tvalues = view(AAPL.values, source_rows, source_cols)
+    
+    APPL2 = TimeArray(tstamps, tvalues, AAPL.colnames, AAPL.meta)
```
iblis17: ha, I got it. `APPL2` is typo

iblis17: It should be `AAPL2`

test/timearray.jl
```diff
@@ -55,6 +55,30 @@ facts("type constructors enforce invariants") do
     end
 end
 
+facts("type constructors allow views") do
+    
+    source_rows = 101:121
+    source_cols = 1:size(AAPL.values)[2]
```
iblis17: please purge the trailing whitespace :)

test/timearray.jl
JeffreySarnoff: :coffee: very good noticing   :grapes:

ararslan:
Looks great, thanks for the contribution!

milktrader:
Should we move away from FactCheck? I guess this is probably an important issue.

JeffreySarnoff:
yes

ararslan:
Yes, definitely. I believe it's no longer actively maintained, plus the Base unit testing framework is fantastic.

iblis17:
@ararslan Here is the blocker I encountered: https://github.com/JuliaQuant/MarketTechnicals.jl/issues/70#issuecomment-298226219

Our testing relies on tons of `roughly`, but the sweet macro is only available in 0.6.

JeffreySarnoff:
roughly is defined using isapprox
```julia

# approx/roughly: Comparing numbers approximately
roughly(x::Number, atol) = (y::Number) -> isapprox(y, x, atol=atol)

roughly(x::Number; kvtols...) = (y::Number) -> isapprox(y, x; kvtols...)

roughly(A::AbstractArray, atol) = (B::AbstractArray) -> begin
    size(A) != size(B) && return false
    return isapprox(A, B, atol=atol)
end

roughly(A::AbstractArray; kvtols...) = (B::AbstractArray) -> begin
    size(A) != size(B) && return false
    return isapprox(A, B; kvtols...)
end
```

Issue #320, opened 2017-07-09
Julia 0.6


femtotrader:
With Julia 0.6 the following warning is displayed

```julia
WARNING: trunc{T <: Real}(x::AbstractArray{T}) is deprecated, use trunc.(x) instead.
Stacktrace:
 [1] depwarn(::String, ::Symbol) at ./deprecated.jl:70
 [2] trunc(::Array{Float64,1}) at ./deprecated.jl:57
 [3] show(::Base.TTY, ::TimeSeries.TimeArray{Float64,2,DateTime,Array{Float64,2}}) at /Users/scls/.julia/v0.6/TimeSeries/src/timearray.jl:73
 [4] print(::Base.TTY, ::TimeSeries.TimeArray{Float64,2,DateTime,Array{Float64,2}}) at ./strings/io.jl:29
 [5] print(::Base.TTY, ::TimeSeries.TimeArray{Float64,2,DateTime,Array{Float64,2}}, ::Char, ::Vararg{Char,N} where N) at ./strings/io.jl:40
 [6] println(::TimeSeries.TimeArray{Float64,2,DateTime,Array{Float64,2}}) at ./coreio.jl:5
 [7] include_from_node1(::String) at ./loading.jl:569
 [8] include(::String) at ./sysimg.jl:14
 [9] process_options(::Base.JLOptions) at ./client.jl:305
 [10] _start() at ./client.jl:371
while loading /Users/femto/.julia/v0.6/DukascopyTicksReader/test/runtests.jl, in expression starting on line 35
```

when running unit tests from https://github.com/femtotrader/DukascopyTicksReader.jl
see https://github.com/femtotrader/DukascopyTicksReader.jl/issues/10

iblis17:
I am working on it :) 
#321

Issue #165, opened 2015-02-13
Refactor inner constructor


milktrader:
This is more for readability. There are currently 8 lines of code that clutter the TimeArray definition that could be take out into another non-exported function called `check_constraints` or something along those lines. 

``` julia
nrow, ncol = size(values, 1), size(values, 2)
nrow != size(timestamp, 1) ? error("values must match length of timestamp"):
ncol != size(colnames,1) ? error("column names must match width of array"):
timestamp != unique(timestamp) ? error("there are duplicate dates"):
~(flipud(timestamp) == sort(timestamp) || timestamp == sort(timestamp)) ? error("dates are mangled"):
flipud(timestamp) == sort(timestamp) ? 
new(flipud(timestamp), flipud(values), colnames):
new(timestamp, values, colnames)
```

iblis17:
can we close this now?

milktrader:
Thanks!

PR #321, opened 2017-07-09
Upgrade to Julia 0.6

iblis17:


src/timearray.jl
```diff
@@ -15,17 +15,15 @@ immutable TimeArray{T, N, D<:TimeType, A<:AbstractArray} <: AbstractTimeSeries
...
-                           #new(timestamp, values, colnames, meta)
+        nrow, ncol = size(values, 1), size(values, 2)
+        nrow != size(timestamp, 1) ? error("values must match length of timestamp"):
+        ncol != size(colnames,1) ? error("column names must match width of array"):
+        # length(unique(colnames)) != ncol ? colnames = replace_dupes(colnames):
+        timestamp != unique(timestamp) ? error("there are duplicate dates"):
+        ~(flipdim(timestamp, 1) == sort(timestamp) || timestamp == sort(timestamp)) ? error("dates are mangled"):
+        flipdim(timestamp, 1) == sort(timestamp) ?
+        new(flipdim(timestamp, 1), flipdim(values, 1), replace_dupes(colnames), meta):
+        new(timestamp, values, replace_dupes(colnames), meta)
```
iblis17: Can I rewrite this single huge conditional ternary statement into if statements for readibility? Any effect on performance if we use `if` ?

ararslan: > Can I rewrite this single huge conditional ternary statement into if statements for readibility?

Oooooh yes, please do. That code is mildly hideous as is.

> Any effect on performance if we use if ?

Nope, ternaries parse as `if` so they're entirely equivalent.
```julia
julia> parse("a ? b : c")
:(if a
        b
    else
        c
    end)
```

iblis17: done in 7a76ac8

src/timearray.jl
```diff
@@ -17,14 +17,21 @@ struct TimeArray{T, N, D <: TimeType, A <: AbstractArray{T, N}} <: AbstractTimeS
...
-        new(flipdim(timestamp, 1), flipdim(values, 1), replace_dupes(colnames), meta):
-        new(timestamp, values, replace_dupes(colnames), meta)
+
+        if nrow != size(timestamp, 1)
+            error("values must match length of timestamp")
+        elseif ncol != size(colnames, 1)
+            error("column names must match width of array")
+        elseif timestamp != unique(timestamp)
+            error("there are duplicate dates")
+        elseif !(flipdim(timestamp, 1) == sort(timestamp) || timestamp == sort(timestamp))
```
ararslan: The latter part of this condition could be written as `issorted(timestamp)`

iblis17: 👌 @ a72cfe6

(going to sleep 💤  , am 2 here)

ararslan: Wow, you're quite productive in the wee hours of the morning. 😄 Thanks for all your great work here!

src/timearray.jl
milktrader: https://github.com/JuliaStats/TimeSeries.jl/issues/165

src/timearray.jl
```diff
@@ -2,30 +2,36 @@
...
-                           new(timestamp, values, replace_dupes(colnames), meta)
-                           #new(timestamp, values, colnames, meta)
+    function TimeArray{T, N, D, A}(
+            timestamp::AbstractVector{D},
+            values::A,
+            colnames::Vector{String},
+            meta::Any) where {T, N, D <: TimeType, A <: AbstractArray{T, N}}
+        nrow, ncol = size(values, 1, 2)
+
+        if nrow != size(timestamp, 1)
```
ararslan: Since `timestamp` is an `AbstractVector`, we know that its size in the first dimension is equal to its length, so this can just be `length(timestamp)` which is a little clearer. The same goes for `colnames` below.

src/timearray.jl
```diff
@@ -2,30 +2,36 @@
...
+            error("values must match length of timestamp")
```
ararslan: Perhaps we should be using typed exceptions here. For example, this and the check below could be a `DimensionMismatch`, duplicate dates could be an `ArgumentError`, etc. If there any `@test_throws` in the tests those will have to be updated accordingly though.

src/timearray.jl
```diff
@@ -2,30 +2,36 @@
...
+        elseif ncol != size(colnames, 1)
+            error("column names must match width of array")
+        elseif timestamp != unique(timestamp)
```
ararslan:
```julia
allunique(timestamp)
```
New in 0.6 🙂

src/timearray.jl
```diff
@@ -2,30 +2,36 @@
...
+            error("there are duplicate dates")
+        elseif !(flipdim(timestamp, 1) == sort(timestamp) || issorted(timestamp))
```
ararslan: `flipdim(x, 1)` is the same as `reverse(x)` if `x` is an `AbstractVector`, so this is just checking whether the dates are sorted in reverse order. Thus the check could be
```julia
!(issorted(timestamp) || issorted(timestamp, rev=true))
```

src/timearray.jl
```diff
@@ -2,30 +2,36 @@
...
+            error("dates are mangled")
+        elseif flipdim(timestamp, 1) == sort(timestamp)
```
ararslan:
```julia
issorted(timestamp, rev=true)
```
And below,
```julia
new(reverse(timestamp), ...)
```
We still need the `flipdim` on `values` though, since that isn't constrained to any particular dimensionality.

src/timearray.jl
milktrader: the history behind this was to avoid `length` in favor of `size` for consistent syntax to follow DataFrames. I think I originally had `length`. In any case, I agree with using `length`.

src/timearray.jl
milktrader: nice

iblis17: done in 881dfac

src/timearray.jl
iblis17: done in 9ffb151

src/timearray.jl
iblis17: done in 9ffb151

src/timearray.jl
iblis17: maybe another PR for this. I want not to change test case if possible in this PR.

src/timearray.jl
iblis17: done in 02dd699

src/apply.jl
```diff
@@ -39,22 +39,36 @@ end # loop
...
             cnames  = [name * string($op) * string(var) for name in ta.colnames]
-            vals = ($op)(ta.values, var)
+            vals = broadcast($op, ta.values, var)
             TimeArray(ta.timestamp, vals, cnames, ta.meta)
         end # function
+
+        # FIXME: make some function allowed dot version only
+        # e.g. `ta > 42` is not allowed, but `ta .> 42` does
+        if Symbol($op) ∉ [$MATH_DOTONLY; $COMPARE_DOTONLY]
+            ($op)(ta::TimeArray{T, N}, var::Number) where {T <: Number, N} =
```
iblis17: I tried to make `ta .> 0` work ; make `ta > 0` raise error. These behaviours are same as before.
But seems defining `broadcast(::typeof(>), ...)` only does not work...

iblis17: In this commit (9237607), 
  - `cl .+ 42` works
  - `cl + 42` works
  - `cl > 0` raise error, this is expected
  - `cl .> 0` raise error, it should broadcast
```julia
julia> cl .> 0
ERROR: MethodError: no method matching isless(::Int64, ::TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}})
Closest candidates are:
  isless(::Real, ::AbstractFloat) at operators.jl:97
  isless(::Real, ::Real) at operators.jl:266
Stacktrace:
 [1] (::##5#6)(::TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}}) at ./<missing>:0
 [2] broadcast(::Function, ::TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}}) at ./broadcast.jl:434
```

any idea?

iblis17: weird... `broadcast(>, cl, 42)` works fine, do I mess up dispatching?

ararslan: Try overloading `broadcast(::typeof(isless), ...)`, since `>` is defined in terms of `<`, which is just an alias for `isless`.

iblis17: hmmm... doesn't work
diff --git a/src/apply.jl b/src/apply.jl
```patch
index e3079a2..13d4fdc 100644
--- a/src/apply.jl
+++ b/src/apply.jl
@@ -55,6 +55,12 @@ for op in [MATH_ALL; COMPARE_DOTONLY]
     end # eval
 end # loop
 
+function broadcast(::typeof(isless), ta::TimeArray{T, N}, var::Number) where {T <: Number, N}
+    cnames = [name * string(<) * string(var) for name in ta.colnames]
+    vals = broadcast(isless, ta.values, var)
+    TimeArray(ta.timestamp, vals, cnames, ta.meta)
+end
+
 # Scalar <--> TimeArray
 for op in [MATH_ALL; COMPARE_DOTONLY]
     @eval begin
```

found that the function list in stacktrace keeping changes:
```julia
julia> isless.(cl, 1)
...
Stacktrace:
 [1] (::##1#2)(::TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}}) at ./<missing>:0
 [2] broadcast(::Function, ::TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}}) at ./broadcast.jl:434

julia> isless.(cl, 1)
...
Stacktrace:
 [1] (::##3#4)(::TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}}) at ./<missing>:0
 [2] broadcast(::Function, ::TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}}) at ./broadcast.jl:434

julia> isless.(cl, 1)
..
Stacktrace:
 [1] (::##5#6)(::TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}}) at ./<missing>:0
 [2] broadcast(::Function, ::TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}}) at ./broadcast.jl:434
```
Is this related to [automatically fusing on dot call](https://docs.julialang.org/en/latest/manual/mathematical-operations/#man-dot-operators-1)?

iblis17:
I just read this blog post about dot-call: https://julialang.org/blog/2017/01/moredots.
I want to deprecate all of the non-dot-call, enforce user to do dot-call on all of element-wise (besides the unary `+` and `-`), in order to get the benefit from syntactic loop fusion, e.g. more performance and broadcast will work for all non-standard function.

And... with this automatically fusion on dot-call, we will lost the information about the original function call, because of Julia compiler creating a single temporary function. So we cannot format our column names as before. Can we accept losing this column name auto formatting feature?

src/timearray.jl
```diff
@@ -64,14 +70,14 @@ isempty(ta::TimeArray) = (length(ta) == 0)
 
 ###### show #####################
 
-function show{T,N}(io::IO, ta::TimeArray{T,N})
+function show(io::IO, ta::TimeArray{T}) where {T}
```
iblis17: Any bad side effect on removing unused type parameter `N`?

ararslan: Nope

ararslan: In particular, the type `TimeArray{T}` should have an implicit `where N`, since the `TimeArray` type has two type parameters, `T` and `N`, and both still exist. That is, it should be the case that `TimeArray{T} where T == TimeArray{T,N} where N where T`.

iblis17:
About the function `basecall`:
Now, part of functionality of `basecall` could be replaced by dot-call.
e.g.: In the second test case (in file `test/apply.jl`):
```julia
@testset "basecall works with Base methods" begin                                                                      
    @testset "cumsum works" begin                                                                                     
        @test basecall(cl, cumsum).values[2] == cl.values[1] + cl.values[2]                                           
    end                                                                                                                
                                                                                                                       
    @testset "log works" begin                                                                                        
        @test basecall(cl, log).values[2] == log(cl.values[2])
    end                                                                                                                
end
```

Note that `basecall(cl, log)` will pop up the deprecated warning.
I'm going to remove this test case and similar cases.

milktrader:
Catching up with these changes (looks good so far) ... are you saying to remove the `basecall` method in favor of something nicer?

iblis17:
> are you saying to remove the `basecall` method in favor of something nicer?

No, `basecall` is still needed some function for which there is no dot-call version, e.g.: `basecall(cl, cumsum)`

But for something like `log`, broadcast via `log.(cl)` will be the better choice.

iblis17:
I think this PR is ready for review.
Although its broadcasting functionality is still not perfect, it can deal with most of cases. Maybe open another PR to make it perfect later. I think this PR is huge enough :p 

FIXME: following broadcast not work currently.
  * `round.(Int, ta)`
  * `myfunc.(sooo, many, args, ta)`

iblis17:
Bump?

src/apply.jl
```diff
@@ -1,180 +1,92 @@
...
+    vals = broadcast(f, vals1, vals2)
+    TimeArray(tstamp, vals, cnames, meta)
+end
+
+function broadcast(f, ta::TimeArray, args...)
+    vals = broadcast(f, ta.values, args...)
+    TimeArray(ta.timestamp, vals, ta.colnames, ta.meta)
+end
+
+# FIXME: How to deal with f(Number, Number, ..., TimeArray)?
```
ararslan: Probably
```julia
function broadcast(f, ns::Number..., ta::TimeArray, args...)
    vals = broadcast(f, ns..., ta.values, args...)
    TimeArray(ta.timestamp, vals, ta.colnames, ta.meta)
end
```
(untested, just a guess)

src/deprecated.jl
```diff
@@ -10,3 +10,32 @@ using Base: @deprecate
...
+         :|, :&, :<, :>, :(==), :(!=), :>=, :<=)
+    @eval import Base: $f
+    @eval @deprecate $f(ta::TimeArray, args...) $f.(ta, args...)
+    @eval @deprecate $f(n::Number, ta::TimeArray) $f.(n, ta)
+end
+
+# non-dot operators
+import Base: $, !, ~
+
+@deprecate ($)(ta1::TimeArray, ta2::TimeArray) xor(ta1, ta2)
```
ararslan: Shouldn't the deprecation replacement be `xor.(ta1, ta2)`?

iblis17: 🆗 @ 7ee48a2

src/apply.jl
iblis17: hmm. `...` can only append to final argument

iblis17: I think we should  do the integration like this: https://github.com/JuliaArrays/StaticArrays.jl/pull/241

Maybe another PR to make it? seems quite complicated

ararslan: Ah right, I forgot about not being able to use `...` for non-final arguments. Never mind then.

iblis17:
Bump?

femtotrader:
Closes #320

PR #324, opened 2017-07-30
using typed exceptions if possible

### LINKED ISSUE #322: Using typed exceptions in TimeArray inner constructor
See https://github.com/JuliaStats/TimeSeries.jl/pull/321#discussion_r126492578

iblis17:
close #322

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,8 @@
+### 0.11.0
+
+* `TimeArray` constructor, `merge`, `update`, `vcat` and `rename`
+  will throw typed exceptions. (issue 322)
+
 ### 0.10.0
 
 * add support for time series plotting via RecipesBase dependency (thank you @mkborregaard) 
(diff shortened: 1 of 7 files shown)


ararslan:
Looks like there are some other `@test_throws` that need to be updated as well.

iblis17:
I thing I can make more exceptions meaningful.

e.g.
   we have the `vcat`: 
     https://github.com/JuliaStats/TimeSeries.jl/blob/master/src/combine.jl#L97-L105
   And the current test cases:
     https://github.com/JuliaStats/TimeSeries.jl/blob/master/test/combine.jl#L147-L161
   Now this PR change some exceptions of inner constructor to type exceptions, and the test cases in `test/combine.jl` line 161 will pop ArgumentError but the other (line 147 line 154) not.
  I think all of combine execptions should switch to ArgumentError.

iblis17:
ready for review :)

PR #335, opened 2017-10-18
timearray: implement `timearray[]`

iblis17:


ararslan:
I'm kind of confused by this. I don't think any other array type behaves this way, does it?

iblis17:
hmm, what i expect is `ta[]` returns the first item of ta, like `Array` in `Base`.

ararslan:
My understanding is that it's actually more of an accident than a feature. In fact, if you trace the calls that are being made in `A[]` (for `A` an array), it comes down to the following definition:
```julia
_to_linear_index(A::AbstractArray) = 1 # TODO: DEPRECATE FOR #14770
```
So hopefully that means the method will actually go away sooner or later.

PR #342, opened 2017-11-07
Migrate Sphinx to Documenter

iblis17:
Also, need help for configuring Travis's auto-deployment. I do not have permission to do that.


ararslan:
I set up the deployment stuff, so that part should be good.

.travis.yml
```diff
@@ -11,3 +11,8 @@ script:
   - julia -e 'Pkg.clone(pwd()); Pkg.build("TimeSeries"); Pkg.test("TimeSeries", coverage=true)'
 after_success:
   - julia -e 'cd(Pkg.dir("TimeSeries")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
+
+  # Documenter auto-deploy
+  - julia -e 'Pkg.add("Documenter")'
+  - julia -e 'Pkg.add("MarketData")'
+  - julia -e 'cd(Pkg.dir("MarketTechnicals", "docs")); include("make.jl")'
```
ararslan: Did you mean TimeSeries rather than MarketTechnicals?

iblis17: oh, sorry, I just copy-paste :p

iblis17:
hmm, pyplot doesn't work ? https://travis-ci.org/JuliaStats/TimeSeries.jl/builds/298389627#L1629

.travis.yml
```diff
@@ -11,3 +11,12 @@ script:
   - julia -e 'Pkg.clone(pwd()); Pkg.build("TimeSeries"); Pkg.test("TimeSeries", coverage=true)'
 after_success:
   - julia -e 'cd(Pkg.dir("TimeSeries")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
+
+  # Documenter auto-deploy
+  - julia -e 'Pkg.add("Documenter"); using Documenter'
+  - julia -e 'Pkg.add("MarketData"); using MarketData'
+  - julia -e 'Pkg.add("Plots")'
+  - julia -e 'Pkg.add("PyCall")'
+  - julia -e 'ENV["PYTHON"]=""; Pkg.build("PyCall"); using PyCall'
```
ararslan: Why is `using` necessary here? These are all run as separate Julia sessions, so `using PyCall` here won't bring PyCall into the namespace in the next step.

ararslan: Also it's probably better to set environment variables in the Travis config, i.e. `env: PYTHON=""` in the appropriate spot.

iblis17: well, I try to use `using` to trigger compilation; hope
it keeps from timeout: https://travis-ci.org/JuliaStats/TimeSeries.jl/builds/298415239#L1715

ararslan: If precompilation is what's eating up all of that time, try deploying from 0.7 instead of 0.6. I believe that's been fixed on master.

iblis17: hmm, we currently do not have 0.7 in our build matrix.
And I want that Documenter renders all `@repl`, `@example` blocks in stable version, and it will not pop up depwarn in documentation.

iblis17:
It conquers the timeout: https://travis-ci.org/JuliaStats/TimeSeries.jl/builds/298428320#L1855
But PyPlot still doesn't work 😂 .
(going to do more experiment on my local box about how to installing PyPlot

iblis17:
hmm, PlotlyJS failed also.... https://travis-ci.org/JuliaStats/TimeSeries.jl/builds/299110732#L1659
Maybe plotting with CI is a bad idea.

iblis17:
it ready for review now: https://travis-ci.org/JuliaStats/TimeSeries.jl/builds/299121037#L1669

.travis.yml
```diff
@@ -21,5 +21,7 @@ after_success:
   - julia -e 'Pkg.add("Plots")'
   - julia -e 'Pkg.add("PyCall")'
   - julia -e 'Pkg.build("PyCall"); using PyCall'
-  - julia -e 'Pkg.add("PyPlot"); using Plots; pyplot()'
+  - julia -e 'Pkg.add("PyPlot")'
```
ararslan: Seems like building PyPlot should happen here rather than after Plots tries to load it.

ararslan: Ah, never mind. I see this is now outdated.

docs/src/timearray.md
```diff
@@ -0,0 +1,55 @@
+# The `TimeArray` time series type
+
+The `TimeArray` time series type is defined here
+(with inner constructor code removed for readability):
+
+```julia
+struct TimeArray{T,N,D<:TimeType,A<:AbstractArray} <: AbstractTimeSeries
+  timestamp::Vector{D}
```
ararslan: Should use the Julia standard of 4 spaces for indentation

iblis17: 👌

iblis17:
autosquashed those `fixup!`s

iblis17:
I will merge this PR tmr if no objection.

femtotrader:
Just a newbie question...
What are the pros and cons of migrating from Sphinx to Documenter ?

iblis17:
@femtotrader oh, Documenter's `@repl` `@example` are quite attractive.
And the second big benefit is that we can use docstring for our functions and render in both of REPL and doc now. (currently we only docs our function in document, no docstring, in other word, we do not have the help function in REPL.)

femtotrader:
Thanks @iblis17 
Do you know if it's possible to test doctests (in an other project) with a Sphinx doc based project?

iblis17:
AFAICT, no implementation for Sphinx based....

iblis17:
@femtotrader I will suggest that doing the migration. At least, Documenter.jl is serving as Julialang official documentation system. docs.julialang.org migrated to it from 0.6 already.

PR #350, opened 2017-12-15
Higher order differences for diff function

iht:
This pull request adds a new parameter `differences` to the `diff` function, to calculate higher order differences. The default value is `1`, so this change is backwards compatible.

### DIFF
--- a/docs/src/apply.md
+++ b/docs/src/apply.md
@@ -60,6 +60,22 @@ put the last observation in the object (which happens to be on Dec 31,
 lead(cl, 499)
 ```
 
+## `diff`
+
+Differentiating a time series calculates the finite difference between
+two consecutive points in the time series. The resulting time series
+will have less points than the original. Those points are filled with
+`NaN` values if `padding=true`.
+
+```@repl diff
+diff(cl)
+```
+
+You can calculate higher order differences by using the keyword
+parameter `differences`, accepting a positive integer. The default
+value is `differences=1`. For instance, passing `differences=2` is
+equivalent to doing `diff(diff(cl)`. 
+
 ## `percentchange`
 
 Calculating change between timestamps is a very common time series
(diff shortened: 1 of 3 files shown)


docs/src/apply.md
```diff
@@ -60,6 +60,22 @@ put the last observation in the object (which happens to be on Dec 31,
...
+`NaN` values if `padding=true`.
+
+```@repl diff
+diff(cl)
+```
+
+You can calculate higher order differences by using the keyword
+parameter `differences`, accepting a positive integer. The default
+value is `differences=1`. For instance, passing `differences=2` is
+equivalent to doing `diff(diff(cl)`.
```
iblis17: :trollface: 
```julia
julia> diff(diff(cl)
                                                                        
ERROR: syntax: incomplete: premature end of input
```

test/apply.jl
```diff
@@ -62,6 +62,22 @@ using TimeSeries
...
+        @test diff(op, padding=true, differences=2).values[3] == diff(op, differences=2).values[1]
+    end
+
+    @testset "diff calculates 3rd-order differences" begin
+        @test diff(op, differences=3).timestamp               == diff(op, padding=false, differences=3).timestamp
+        @test diff(op, differences=3).values                  == diff(op, padding=false, differences=3).values
+        @test diff(diff(diff(op))).timestamp                  == diff(op, padding=false, differences=3).timestamp
+        @test diff(diff(diff(op))).values                     == diff(op, padding=false, differences=3).values
+        @test diff(op, padding=true, differences=3).values[4] == diff(op, differences=3).values[1]
+    end
```
iblis17: maybe add some test cases for multi-column TimeArray `ohlc`?

iht: Ok, I will push a new commit with more tests.

iblis17:
Thanks for your contribution!

PR #352, opened 2017-12-19
Add copy function to TimeArray

iht:
The copy base function is necessary to use the [Bootstrap package](https://github.com/juliangehring/Bootstrap.jl).

The Bootstrap package will also need some changes, but this paves the way to being able to use both packages together.

### DIFF
--- a/src/timearray.jl
+++ b/src/timearray.jl
@@ -1,7 +1,7 @@
 ###### type definition ##########
 
-import Base: convert, length, show, getindex, start, next, done, isempty, endof,
-             size, eachindex
+import Base: convert, copy, length, show, getindex, start, next, done, isempty,
+             endof, size, eachindex
 
 abstract type AbstractTimeSeries end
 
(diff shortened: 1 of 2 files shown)


src/timearray.jl
```diff
@@ -56,6 +56,11 @@ convert(::Type{TimeArray{Float64, 2}}, x::TimeArray{Bool, 2}) =
 convert(x::TimeArray{Bool, 1}) = convert(TimeArray{Float64, 1}, x::TimeArray{Bool, 1})
 convert(x::TimeArray{Bool, 2}) = convert(TimeArray{Float64, 2}, x::TimeArray{Bool, 2})
 
+###### copy ###############
+
+copy(ta::TimeArray)::TimeArray =
+         TimeArray(ta.timestamp, ta.values, ta.colnames, ta.meta)
```
iblis17: Could you make it 4-space indented?

src/timearray.jl
```diff
@@ -56,6 +56,11 @@ convert(::Type{TimeArray{Float64, 2}}, x::TimeArray{Bool, 2}) =
 convert(x::TimeArray{Bool, 1}) = convert(TimeArray{Float64, 1}, x::TimeArray{Bool, 1})
 convert(x::TimeArray{Bool, 2}) = convert(TimeArray{Float64, 2}, x::TimeArray{Bool, 2})
 
+###### copy ###############
+
+copy(ta::TimeArray)::TimeArray =
```
iblis17: I think our compiler is smart enough, and we do not need return type annotation here.

iht: Even if it is redundant, I prefer to annotate types as much as possible. But I will adapt to the style you prefer. Would you prefer to remove the output type?

iblis17: ok, I'm fine with current.

iht:
Thanks!

PR #357, opened 2017-12-25
timearray: Revoke deprecation of `==` and redefine its meaning

iblis17:
```julia
julia> cl == copy(cl)
true
```

---

@dourouc05 please review this

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,10 @@
 ### 0.12.0
 
 * Revoking deprecation warning of `==` and redefining its meaning as
-  'comparing all fields of two TimeArray'. (#356, #357)
+  'comparing all fields of two `TimeArray`s'.
+  Note that if two `TimeArray`s have different dimension, we consider that is
+  unequal.
+  (#356, #357)
 
   ```julia
   julia> cl == copy(cl)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,13 @@
+### 0.12.0
+
+* Revoking deprecation warning of `==` and redefining its meaning as
+  'comparing all fields of two TimeArray'. (#356, #TBD)
+
+  ```julia
+  julia> cl == copy(cl)
+  true
+  ```
+
 ### 0.11.0
 
 * Dropping 0.5 support. (issue [#327])
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,12 +1,24 @@
 ### 0.12.0
 
 * Revoking deprecation warning of `==` and redefining its meaning as
-  'comparing all fields of two TimeArray'. (#356, #TBD)
+  'comparing all fields of two TimeArray'. (#356, #357)
 
   ```julia
   julia> cl == copy(cl)
   true
   ```
+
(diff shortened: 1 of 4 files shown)


dourouc05:
It works fine for me! (I have never really played with generated functions, so don't count me in for the code review.)

Maybe the docstring could me more informative for users? I am not sure beginners will know what fields a `TimeArray` has. This is what I propose: 

> If true, all fields of x and y should be equal, meaning that the two time arrays have the same values at the same points in time, and the same metadata.

iblis17:
ok, I updated it

src/timearray.jl
```diff
@@ -69,13 +69,67 @@ length(ata::AbstractTimeSeries) = length(ata.timestamp)
...
+Implies
+
+```julia
+x.timestamp == y.timestamp &&
+x.values    == y.values    &&
+x.colnames  == y.colnames  &&
+x.meta      == y.meta
+```
+"""
+@generated function ==(x::TimeArray{T,N}, y::TimeArray{S,M}) where {T,S,N,M}
```
ararslan: This shouldn't require a generated function.
```julia
Base.:(==)(x::TimeArray{T,N}, y::TimeArray{S,N}) where {T,S,N} =
    all(f->getfield(x, f) == getfield(y, f), fieldnames(TimeArray))
Base.:(==)(x::TimeArray{T,N}, y::TimeArray{S,M}) where {T,S,N,M} = false
```

src/timearray.jl
```diff
@@ -69,13 +69,67 @@ length(ata::AbstractTimeSeries) = length(ata.timestamp)
...
+        for f ∈ fieldnames(TimeArray)
+          !isequal(getfield(x, f), getfield(y, f)) && return :false
+        end
+        :true
+    end
+end
+
+# support for Dict
+hash(x::TimeArray, h::UInt) =
+    mapreduce(+, fieldnames(TimeArray)) do f
```
ararslan: `sum` is `mapreduce(+, ...)` so this can be written
```julia
Base.hash(x::TimeArray, h::UInt) = sum(f->hash(getfield(x, f), h), fieldnames(TimeArray))
```

NEWS.md
```diff
@@ -1,3 +1,25 @@
+### 0.12.0
+
+* Revoking deprecation warning of `==` and redefining its meaning as
+  'comparing all fields of two TimeArray'. (#356, #357)
```
ararslan: It would also be good to note here that two `TimeArray`s compare unequal if they have different dimensions.

src/timearray.jl
iblis17: Wow, it's elegant.
TIL, `all` will do early stopping.

iblis17: and done in following commits

src/timearray.jl
iblis17: done

NEWS.md
iblis17: done

iblis17:
good to go?

PR #363, opened 2018-03-26
merge() optimization

ancapdev:
Summary:
* Unchecked construction of TimeArray objects.
* Direct implementation of outer join.
* Support for arbitrary types by user defined missing value.
* More optimal index based assignment than generic indexing operation (applies to left, right, and outer join).

Mostly focused on optimizing outer join. Benchmark example:

```Julia
using TimeSeries
using BenchmarkTools

struct SimpleTime <: Dates.TimeType
    value::Int64
end

Base.isless(x::SimpleTime, y::SimpleTime) = x.value < y.value
Base.isequal(x::SimpleTime, y::SimpleTime) = x.value == y.value
Base.:(==)(x::SimpleTime, y::SimpleTime) = x.value == y.value

t1 = [SimpleTime(x) for x in 1:2_000_000]
t2 = [SimpleTime(2 * x) for x in 1:2_000_000]
v1 = rand(Float64, length(t1), 10)
v2 = rand(Float64, length(t2), 10)
ts1 = TimeArray(t1, v1)
ts2 = TimeArray(t2, v2)
@btime merge(ts1, ts2, :outer)
```

Previous result:
```
  2.008 s (659 allocations: 1.97 GiB)
```

New result:
```
  415.149 ms (518 allocations: 503.57 MiB)
```

At these scales it's mostly load/store bound so any reduction in that (e.g. smaller index types, smaller data types, fewer passes) make the biggest difference. For my, and maybe other people's use cases ```Float32``` support helps a lot. The above benchmark with ```Float32``` values is ~224ms.

### DIFF
--- a/src/combine.jl
+++ b/src/combine.jl
@@ -2,6 +2,14 @@ import Base: merge, hcat, vcat, map
 
 ###### merge ####################
 
+function _merge_outer(::Type{IndexType}, ta1::TimeArray{T, N, D}, ta2::TimeArray{T, M, D}, missingvalue, meta) where {IndexType, T, N, M, D}
+    timestamps, new_idx1, new_idx2 = sorted_unique_merge(IndexType, ta1.timestamp, ta2.timestamp)
+    vals = fill(convert(T, missingvalue), (length(timestamps), length(ta1.colnames) + length(ta2.colnames)))
+    insertbyidx!(vals, ta1.values, new_idx1)
+    insertbyidx!(vals, ta2.values, new_idx2, size(ta1.values, 2))
+    TimeArray(timestamps, vals, [ta1.colnames; ta2.colnames], meta; unchecked = true)
+end
+
 function merge(ta1::TimeArray{T, N, D}, ta2::TimeArray{T, M, D}, method::Symbol=:inner;
                colnames::Vector=[], meta::Any=Void, missingvalue=NaN) where {T, N, M, D}
 
@@ -35,11 +43,11 @@ function merge(ta1::TimeArray{T, N, D}, ta2::TimeArray{T, M, D}, method::Symbol=
 
     elseif method == :outer
 
-        timestamps, new_idx1, new_idx2 = sorted_unique_merge(ta1.timestamp, ta2.timestamp)
-        vals = fill(convert(T, missingvalue), (length(timestamps), length(ta1.colnames) + length(ta2.colnames)))
-        insertbyidx!(vals, ta1.values, new_idx1)
-        insertbyidx!(vals, ta2.values, new_idx2, size(ta1.values, 2))
-        ta = TimeArray(timestamps, vals, [ta1.colnames; ta2.colnames], meta; unchecked = true)
+        ta = if (length(ta1.timestamp) + length(ta2.timestamp)) > typemax(Int32)
+            _merge_outer(Int64, ta1, ta2, missingvalue, meta)
+        else
+            _merge_outer(Int32, ta1, ta2, missingvalue, meta)
+        end
 
     else
         throw(ArgumentError(
--- a/src/combine.jl
+++ b/src/combine.jl
@@ -35,10 +35,8 @@ function merge(ta1::TimeArray{T, N, D}, ta2::TimeArray{T, M, D}, method::Symbol=
 
     elseif method == :outer
 
-        timestamps = sorted_unique_merge(ta1.timestamp, ta2.timestamp)
+        timestamps, new_idx1, new_idx2 = sorted_unique_merge(ta1.timestamp, ta2.timestamp)
         vals = fill(convert(T, missingvalue), (length(timestamps), length(ta1.colnames) + length(ta2.colnames)))
-        new_idx1 = sorted_subset_idx(timestamps, ta1.timestamp)
-        new_idx2 = sorted_subset_idx(timestamps, ta2.timestamp)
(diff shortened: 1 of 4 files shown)


src/utilities.jl
```diff
@@ -25,36 +25,85 @@ function overlaps(ts::Vararg{Vector, N}) where {N}
...
+function sorted_unique_merge(::Type{IndexType}, a::Vector, b::Vector) where {IndexType}
+    # NOTE: Since merge is mostly load/store bound, the lower memory pressure from 32 bit indexing does help.
+    #       Would need to be reconsidered if datasets are larger, or branch the algorithm
+    i, na, j, nb = 1, length(a), 1, length(b)
+    c = similar(a, length(a) + length(b))
+    idx_a = Vector{IndexType}(length(a))
+    idx_b = Vector{IndexType}(length(b))
+    k = 1
+    @inbounds while (i <= na) && (j <= nb)
+		if a[i] < b[j]
```
iblis17: The indentation looks weird here. Please use space.

ancapdev: Think the original code had some tabs and my editor was set to auto. Should be fixed.

src/combine.jl
```diff
@@ -2,8 +2,16 @@ import Base: merge, hcat, vcat, map
...
+    timestamps, new_idx1, new_idx2 = sorted_unique_merge(IndexType, ta1.timestamp, ta2.timestamp)
+    vals = fill(convert(T, missingvalue), (length(timestamps), length(ta1.colnames) + length(ta2.colnames)))
+    insertbyidx!(vals, ta1.values, new_idx1)
+    insertbyidx!(vals, ta2.values, new_idx2, size(ta1.values, 2))
+    TimeArray(timestamps, vals, [ta1.colnames; ta2.colnames], meta; unchecked = true)
+end
+
 function merge(ta1::TimeArray{T, N, D}, ta2::TimeArray{T, M, D}, method::Symbol=:inner;
-               colnames::Vector=[], meta::Any=Void) where {T, N, M, D}
+               colnames::Vector=[], meta::Any=Void, missingvalue=NaN) where {T, N, M, D}
```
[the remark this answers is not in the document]
ancapdev: Renamed to ```padvalue```

ancapdev:
@iblis17 Do these look in a state you're happy to merge now?

src/utilities.jl
```diff
@@ -25,36 +25,83 @@ function overlaps(ts::Vararg{Vector, N}) where {N}
...
+end
 
-    return c
+"""
+    insertbyidx!(dst, src, dstidx, srcidx)
 
-end  # sorted_unique_merge
+For each column in src, insert elements from src[srcidx[i], column] to dst[dstidx[i], column].
+"""
+function insertbyidx!(dst::AbstractArray, src::AbstractArray, dstidx::Vector, srcidx::Vector)
```
iblis17: I think this function can replace by 
```julia
broadcast_setindex!(dst, broadcast_getindex(src, srcidx), dstidx)
```

which is faster in your use case?

ancapdev:
```Julia
using BenchmarkTools
using TimeSeries
dst = zeros(1_000_000)
src = zeros(1_000_000)
dstidx = [1:length(dst)...]
srcidx = [1:length(dst)...]
@btime broadcast_setindex!(dst, broadcast_getindex(src, srcidx), dstidx)
@btime TimeSeries.insertbyidx!(dst, src, dstidx, srcidx)
```
  9.486 ms (8 allocations: 7.63 MiB)
  2.208 ms (0 allocations: 0 bytes)

ancapdev: I would love for the core language and library features to work optimally, so worth revisiting this in the future.

iblis17:
Thanks for your great contributions! :+1:

ancapdev:
It's fun to help a little bit when so much great work is done by other people before me :smiley:.

Btw, ```broadcast_setindex!(dst, broadcast_getindex(src, srcidx), dstidx)``` doesn't seem to work multidimensionally, only copying the first column. ```dst[dstidx, :] = src[srcidx, :]``` works of course, and is about 3-4x slower when copying half the rows from a 10_000_000 x 10 array.

iblis17:
Actually, `dst[dstidx, :] = src[srcidx, :]` can have a nice result, and don't benchmark against global variable (Type of global variable is unpredictable, so it isn't being optimized).
```julia
julia> f = (dst, src, srcidx, dstidx) -> @inbounds(dst[dstidx, :] = @view(src[srcidx, :]))                                                         
(::#26) (generic function with 1 method)                                                                                                           
                                                                                                                                                   
julia> @btime f($dst, $src, $srcidx, $dstidx)                                                                                                      
  1.700 ms (5 allocations: 192 bytes)                                                                                                              
                                                                                                                                                   
julia> @btime TimeSeries.insertbyidx!($dst, $src, $dstidx, $srcidx)                                                                                
  1.632 ms (0 allocations: 0 bytes)                                                                                                                
                                                                                                                                                   
julia> @btime broadcast_setindex!($dst, broadcast_getindex($src, $srcidx), $dstidx)                                                                
  4.636 ms (4 allocations: 7.63 MiB)
```

ancapdev:
Yep, aware of the globals type instability. In this case I figured it wasn't going to make much difference because it only affects the dispatch, and the functions in the benchmark operate on a fairly large dataset.

You're right though, with ```@inbounds``` and ```@view``` the performance is about the same, so this is a nicer solution.

PR #370, opened 2018-07-23
Julia 0.7

iblis17:
TODO:

- [x] fix broadcasting 
  https://docs.julialang.org/en/v1/manual/interfaces/#man-interfaces-broadcasting-1
- [x] `Base.findall` support
- [x] New iteration protocol


.travis.yml
```diff
@@ -2,15 +2,16 @@ language: julia
...
 env:
   - PYTHON=""
 notifications:
   email: false
 sudo: false
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  - julia -e 'Pkg.clone(pwd()); Pkg.build("TimeSeries"); Pkg.test("TimeSeries", coverage=true)'
+  - julia -e 'Pkg.clone(pwd()); Pkg.build("TimeSeries")'
+  - julia -e 'Pkg.test("TimeSeries", coverage=true)'
```
ararslan: For compatibility with 0.7/1.0, you should just remove or comment out the `script` block entirely. That allows Travis to use its default script for Julia, which is set up to work with all existing versions.

iblis17: oh, okay.

femtotrader:
I think https://github.com/JuliaStats/TimeSeries.jl/blob/b0773e8ce5bdf38d18c5faf5d22b5dcf50134c11/test/runtests.jl#L13

should be removed according

https://travis-ci.org/JuliaStats/TimeSeries.jl/builds/426145131#L673

src/TimeSeries.jl
```diff
@@ -19,12 +19,12 @@ export TimeArray, AbstractTimeSeries,
...
+include("utilities.jl")
 include("split.jl")
 include("apply.jl")
 include("broadcast.jl")
 include("combine.jl")
 include("readwrite.jl")
-include("utilities.jl")
 include("modify.jl")
 include("basemisc.jl")
 include("deprecated.jl")
```
femtotrader: why keeping and empty `deprecated.jl` file.

https://github.com/JuliaStats/TimeSeries.jl/blob/4fbedc3f4c162e09600d19d5fb2d633f3c8a1aee/src/deprecated.jl

iblis17: I want to keep `src/deprecated.jl` for future use, but I do not want to test the deprecation binding in the future, since it mess up the testing output. I just removed the `test/deprecated.jl`

src/timearray.jl
```diff
@@ -1,9 +1,9 @@
 ###### type definition ##########
 
-import Base: convert, copy, length, show, getindex, start, next, done, isempty,
-             endof, size, eachindex, ==, isequal, hash
+import Base: convert, copy, length, show, getindex, iterate,
+             lastindex, size, eachindex, ==, isequal, hash, ndims
 
-abstract type AbstractTimeSeries end
+abstract type AbstractTimeSeries{T,N,D} end
```
iblis17: A notable change: I make this abstract type being parametric, to help some iteration stuffs like `eltype`.

Issue #294, opened 2016-10-25
head, tail, first, last


milktrader:
Default `head` and `first` methods to return 6 values. Introduce `first`, `last` as single values representing their eponymous values.

milktrader:
Current state of this ...

````julia
julia> head(cl)1x1 TimeSeries.TimeArray{Float64,2,Date,Array{Float64,2}} 2000-01-03 to 2000-01-03

             Close     
2000-01-03 | 111.94    


julia> tail(cl)
1x1 TimeSeries.TimeArray{Float64,2,Date,Array{Float64,2}} 2001-12-31 to 2001-12-31
             Close    
2001-12-31 | 21.9     


julia> first(cl)
(2000-01-03,[111.94])

julia> last(cl)
1x1 TimeSeries.TimeArray{Float64,1,Date,Array{Float64,1}} 2001-12-31 to 2001-12-31

             Close    
2001-12-31 | 21.9     
````

brilhana:
Currently working on this. Will commit later this week.

PR #371, opened 2018-07-28
Added first and last (#294)

brilhana:
Added `first` and `last` methods. `head` and `tail` now defaults to `n = 6`.

### DIFF
--- a/src/TimeSeries.jl
+++ b/src/TimeSeries.jl
@@ -7,7 +7,8 @@ using Base.Dates
 using RecipesBase
 
 export TimeArray, AbstractTimeSeries,
-       when, from, to, findwhen, find, timestamp, values, colnames, meta, head, tail,
+       when, from, to, findwhen, find, timestamp, values, colnames, meta,
+       head, tail, first, last,
        lag, lead, diff, percentchange, moving, upto,
        uniformspaced, uniformspace, dropnan,
        basecall,
--- a/src/TimeSeries.jl
+++ b/src/TimeSeries.jl
@@ -7,8 +7,7 @@ using Base.Dates
 using RecipesBase
 
 export TimeArray, AbstractTimeSeries,
-       when, from, to, findwhen, find, timestamp, values, colnames, meta,
-       head, tail, first, last,
+       when, from, to, findwhen, find, timestamp, values, colnames, meta, head, tail,
        lag, lead, diff, percentchange, moving, upto,
        uniformspaced, uniformspace, dropnan,
        basecall,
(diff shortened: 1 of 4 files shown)


src/split.jl
```diff
@@ -31,21 +31,38 @@ findwhen(ta::TimeArray{Bool, 1}) = ta.timestamp[find(ta.values)]
...
     ncol          = length(ta.colnames)
     tail_start = length(ta)-n+1
     new_timestamp = ta.timestamp[tail_start:end]
     new_values    = ta.values[tail_start:end, 1:ncol]
     TimeArray(new_timestamp, new_values, ta.colnames, ta.meta)
 end
 
+###### first, last ###########
+
+function first(ta::TimeArray)
```
iblis17: Maybe we should overload `Base.first` and `Base.last`?

iblis17: and I think this function should reuse `head`
```julia
Base.first(ta::TimeArray)  = head(ta, 1)
Base.last(ta::TimeArray)  = tail(ta, 1)
```

brilhana: Didn't think about it, good idea.

src/TimeSeries.jl
```diff
@@ -7,7 +7,8 @@ using Base.Dates
 using RecipesBase
 
 export TimeArray, AbstractTimeSeries,
-       when, from, to, findwhen, find, timestamp, values, colnames, meta, head, tail,
+       when, from, to, findwhen, find, timestamp, values, colnames, meta,
+       head, tail, first, last,
```
iblis17: So, I think we do not need to export them now.

test/split.jl
```diff
@@ -55,52 +55,69 @@ end
...
-        @test head(ohlc).meta         == "AAPL"
-
-        @test length(tail(ohlc))      == 1
-        @test length(tail(ohlc,1))    == 1
-        @test tail(ohlc).timestamp[1] == Date(2001,12,31)
-        @test tail(ohlc).values       == [22.51  22.66  21.83  21.9]
-        @test tail(ohlc).meta         == "AAPL"
+    @testset "head, tail, first and last methods work with default n value on multi column TimeArray" begin
+        @test length(head(ohlc)) == 6
+        @test head(ohlc,1)       == [104.88  112.5  101.69  111.94]
```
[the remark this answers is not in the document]
brilhana: Indeed. Fixed it now.

brilhana:
Some tests fail, I'm fixing them now.

iblis17:

PR #377, opened 2018-10-06
Symbol column indexing

iblis17:
ref #262 

- [x] add deprecation
- [x] support `getproperty`
- [x] support `Base.propertynames`
- [x] Update doc


src/utilities.jl
```diff
@@ -115,3 +115,61 @@ end  # setcolnames!
...
+        n += 1
+    end
+    cnames
+end
+
+# helper method for inner constructor
+find_dupes_index(A::Vector{Symbol}) =
+    @inbounds [i for i in eachindex(A) if A[i] ∈ A[1:i-1]]
+
+@memoize function gen_colnames(n::Integer)
```
ararslan: Have you seen a notable performance improvement when using `@memoize`?

iblis17: yeah, this function is quite slow. (or my algo is bad, is there better algo?

Orig:
```
julia> @btime TimeSeries.gen_colnames(5)
  899.902 ns (9 allocations: 384 bytes) 
5-element Array{Symbol,1}:
 :A
 :B
 :C
 :D
 :E
```
vs `@memoize`
```
julia> @btime TimeSeries.gen_colnames(5)
  92.643 ns (2 allocations: 32 bytes)
5-element Array{Symbol,1}:
 :A
 :B
 :C
 :D
 :E
```

test/combine.jl
```diff
@@ -43,30 +43,17 @@ end
...
-        @test merge(cl, op, :outer, colnames=["a","b"]).colnames == ["a", "b"]
-        @test_throws ErrorException merge(cl, op, :outer, colnames=["a"])
-        @test_throws ErrorException merge(cl, op, :outer, colnames=["a","b","c"])
+        @test merge(cl, ohlc[:High, :Low], colnames=[:a, :b, :c]).colnames == [:a, :b, :c]
+        @test merge(cl, op, colnames=[:a, :b]).colnames                    == [:a, :b]
+        @test_throws ErrorException merge(cl, op, colnames=[:a])
+        @test_throws ErrorException merge(cl, op, colnames=[:a, :b, :c])
+
+        for mode ∈ [:inner, :left, :right, :outer]
+          @test merge(cl, ohlc[:High, :Low], mode, colnames=[:a, :b, :c]).colnames == [:a, :b, :c]
```
ararslan: Should be the Julia standard 4 spaces for indentation

src/utilities.jl
ararslan: Interesting. I'll take a look at the algorithm itself in more detail later, this is fine in the meantime I suppose.

test/combine.jl
iblis17: fixed

iblis17:
ready for review.

femtotrader:
LGTM

femtotrader:
But doc should also be updated accordingly

femtotrader:
You might search for `"`  in

https://github.com/JuliaStats/TimeSeries.jl/blob/master/docs/src/indexing.md
https://github.com/JuliaStats/TimeSeries.jl/blob/master/docs/src/modify.md
https://github.com/JuliaStats/TimeSeries.jl/blob/master/docs/src/plotting.md
https://github.com/JuliaStats/TimeSeries.jl/blob/master/docs/src/split.md
https://github.com/JuliaStats/TimeSeries.jl/blob/master/docs/src/combine.md

iblis17:
Ah, sure.
But I want to make another PRs for doc stuffs, including sort out docstring from current doc pages.

femtotrader:
As API is changing (access column using Symbol instead of String) I think it's always safer to document changes and after (in an other PR) improve doc.
Because we don't know how long it will last to create the second PR... and during this delay doc and code will be inconsistent.
That's just my opinion. But code LGTM! You did a great work!

src/timearray.jl
```diff
@@ -331,29 +318,12 @@ lastindex(ta::TimeArray) = length(ta.timestamp)
...
-            else
-                cnames[d] = string(cnames[d][1:length(cnames[d])-length(string(n))-1], "_$n")
-            end
-        end
-        n += 1
+@generated function getproperty(ta::T, c::Symbol) where {T<:AbstractTimeSeries}
+    fs = fieldnames(T)
+
+    quote
+        (c ∈ $fs) ? getfield(ta, c) : ta[c]
```
ararslan: If a user tries to call a column `values`, for example, this will be incredibly odd behavior. Either replace all field accesses within this package to use `getfield` explicitly, or don't support `getproperty`. (I'd vote for the latter.)

src/utilities.jl
```diff
@@ -115,3 +115,75 @@ end  # setcolnames!
...
+# helper method for inner constructor
+function replace_dupes(cnames::Vector{Symbol})
+    n = 1
+    while !allunique(cnames)
+        ds = find_dupes_index(cnames)
+        for d in ds
+            if n == 1
+                cnames[d] = Symbol(cnames[d], "_$n")
+            else
+                s = cnames[d] |> string
```
ararslan: Why change the function call to a pipe? It just makes it harder to read.

src/utilities.jl
```diff
@@ -115,3 +115,75 @@ end  # setcolnames!
...
+
+# helper method for `getindex`
+"""
+Return the first index of the given column names.
+Raise `KeyError` if col name not found.
+"""
+findcol(ta::AbstractTimeSeries, s::Symbol) = findcol(colnames(ta), s)
+
+@inline function findcol(cols::Vector{Symbol}, s::Symbol)
+  i = findfirst(isequal(s), cols)
```
ararslan: Incorrect indentation

test/timearray.jl
```diff
@@ -494,4 +504,20 @@ end
 end  # @testset "show methods don't throw errors"
 
 
+@testset "getproperty" begin
+    let
+        @test cl.Close  == cl[:Close]
+        @test_throws KeyError cl.NotFound
```
ararslan: The type of error should be consistent with what's used by DataFrames

iblis17: yeah, that's why I throw `KeyError`
```julia
julia> using DataFrames
                                                                                                                                                                             
julia> df = DataFrame()
0×0 DataFrame                                                                                                                                                                
                                                                                                                                                                             
                                                                                                                                                                             
julia> df[:A]
ERROR: KeyError: key :A not found                                                                                                                                            
Stacktrace:                                                                                                                                                                  
 [1] getindex at ./dict.jl:478 [inlined]                                                                                                                                     
 [2] getindex at /home/iblis/.julia/packages/DataFrames/1PqZ3/src/other/index.jl:123 [inlined]                                                                               
 [3] getindex(::DataFrame, ::Symbol) at /home/iblis/.julia/packages/DataFrames/1PqZ3/src/dataframe/dataframe.jl:264                                                          
 [4] top-level scope at none:0                                                                                                                                               
                                                                                                                                                                             
julia> df.A
ERROR: KeyError: key :A not found                                                                                                                                            
Stacktrace:                                                                                                                                                                  
 [1] getindex at ./dict.jl:478 [inlined]                                                                                                                                     
 [2] getindex at /home/iblis/.julia/packages/DataFrames/1PqZ3/src/other/index.jl:123 [inlined]                                                                               
 [3] getindex at /home/iblis/.julia/packages/DataFrames/1PqZ3/src/dataframe/dataframe.jl:264 [inlined]                                                                       
 [4] getproperty(::DataFrame, ::Symbol) at /home/iblis/.julia/packages/DataFrames/1PqZ3/src/abstractdataframe/abstractdataframe.jl:241
```

ararslan: Ah, I see. For some reason I thought DataFrames did something else. Sorry about that.

src/timearray.jl
ararslan: Yeah, every use of `ta.values` would need to be changed to `getfield(ta, :values)`.

iblis17: I will change them to our helper function `vaules(ta)`, `timestamp(ta)`, `colnames(ta)` and `meta(ta)`

iblis17:
Rebased. good to go?

docs/src/modify.md
```diff
@@ -28,7 +28,7 @@ The `rename` method allows the column name(s) to be changed:
 ```@repl
 using TimeSeries
 using MarketData
-rename(cl, "New Close")
-rename(cl, ["New Close"])
-rename(ohlc, ["New Open", "New High", "New Low", "New Close"])
+rename(cl, :Close′)
```
ararslan: Is the `'` here intentional?

docs/src/timearray.md
```diff
@@ -48,8 +48,18 @@ where `n` enumerates from 1.
...
+type `Nothing`. This default is designed to allow programmers to ignore
 this field. For those who wish to utilize this field, `meta` can hold
 common types such as `String` or more elaborate user-defined types. One
 might want to assign a name to an object that is immutable versus
 relying on variable bindings outside of the object's type fields.
+
+## Fields getter functions
+
+There are four field getter functions exported.
+They are named as same as the field names.
```
ararslan: Field names are effectively an implementation detail, so I don't know that it's worth mentioning that. Good to document these functions though.

src/combine.jl
```diff
@@ -2,48 +2,55 @@ import Base: merge, hcat, vcat, map
...
-        meta = ta1.meta
-    elseif typeof(ta1.meta) <: AbstractString && typeof(ta2.meta) <: AbstractString && meta isa Nothing
-        meta = string(ta1.meta, "_", ta2.meta)
+    if colnames isa Vector{<:AbstractString}
+        @warn "`merge(...; colname::Vector{<:AbstractString})` is deprecated, " *
+              "use `merge(...; colnames=Symbol.(colnames))` instead."
+        colnames = Symbol.(colnames)
+    end
+
+    meta = if _meta(ta1) == _meta(ta2) && meta isa Nothing
```
ararslan: `meta === nothing` is typically preferred to `meta isa Nothing`. In particular, `===` makes the compiler happy. (Though `isa` is still better than `==` in this case.)

src/combine.jl
```diff
@@ -2,48 +2,55 @@ import Base: merge, hcat, vcat, map
...
+        _meta(ta1)
+    elseif typeof(_meta(ta1)) <: AbstractString && typeof(_meta(ta2)) <: AbstractString && meta isa Nothing
+        string(_meta(ta1), "_", _meta(ta2))
     else
-        meta = meta
+      meta
```
ararslan: Indentation

src/utilities.jl
```diff
@@ -115,3 +115,75 @@ end  # setcolnames!
...
+# helper method for inner constructor
+function replace_dupes(cnames::Vector{Symbol})
+    n = 1
+    while !allunique(cnames)
+        ds = find_dupes_index(cnames)
+        for d in ds
+            if n == 1
+                cnames[d] = Symbol(cnames[d], "_$n")
+            else
+              s = string(cnames[d])
```
ararslan: Indentation

src/utilities.jl
ararslan: IIRC it's just `x` appended with sequential numbers, i.e. `x1`, `x2`, ...

ararslan:
Looks good to me for the most part, just a few comments. Nice work here!

docs/src/modify.md
femtotrader: `′` (not `'`) is used several times in https://github.com/JuliaStats/TimeSeries.jl/blob/master/src/broadcast.jl

I was surprised by the use of such a character also which can't be found easily on my keyboard

ararslan: Those render exactly the same for me. Weird.

femtotrader: Why not naming simply `:NewClose` (or `:Close2`)?

What are the english name of `′`, `'` and `

femtotrader: See screenshot to see how it's rendering here (France)
capture d ecran 2018-10-16 a 22 39 24

ararslan: Not sure about the first one, the second is "apostrophe," and the third is "backtick" (assuming that's rendering for me properly).

iblis17: > Why not naming simply `:NewClose` (or `:Close2`)?

well, I personally love the unicode support in Julia.
I always write code (in my research project) like `x = 1; x\prime<tab> = x + 42`
or 
`x\vec<tab> ... `

```julia
julia> x = 1
1

julia> x′ = x + 42
43

julia> function ϕ()
        42
       end 
ϕ (generic function with 1 method)

julia> :Close₂                                                   
```

src/combine.jl
iblis17: ha, okay 5d1dcc7

src/combine.jl
iblis17: 9262ce7

src/utilities.jl
iblis17: 9262ce7

docs/src/timearray.md
iblis17: I think the field name is revealed to user already here:
http://juliastats.github.io/TimeSeries.jl/latest/timearray.html

So I add that section.

iblis17:
One more thing to be discussed:
This warning will pop up during recompilation. Is it safe to ignore it ?

```julia
julia> using TimeSeries
[ Info: Recompiling stale cache file /home/iblis/.julia/compiled/v1.0/TimeSeries/dmqCl.ji for TimeSeries [9e3dc215-6440-5c97-bce1-76c03772f85e]
WARNING: eval from module Memoize to TimeSeries:
Expr(:const, Symbol("##gen_colnames_memoized_cache") = Expr(:call, :IdDict))
  ** incremental compilation may be broken for this module **

WARNING: eval from module Memoize to TimeSeries:
Expr(:const, Symbol("##carry_char_memoized_cache") = Expr(:call, :IdDict))
  ** incremental compilation may be broken for this module **
```

docs/src/modify.md
ararslan: Oh, I see. I was on Linux before when it rendered like an apostrophe for me, but on macOS they're distinct.

femtotrader: @ararslan maybe you should open an issue on your Linux distribution bug tracker because that's quite problematic to have 2 differents characters rendered the same.

iblis17:
I decided to drop Memoize.jl finally

PR #379, opened 2018-10-24
docs: add docstring for TimeArray field getters

iblis17:


src/timearray.jl
```diff
@@ -326,13 +326,47 @@ propertynames(ta::TimeArray) = colnames(ta)
...
+Get the user-defined metadata of a `TimeArray`.
+"""
+meta(ta::TimeArray) = getfield(ta, :meta)
 
 # internal use, to avoid name collision
 _timestamp(ta::TimeArray) = getfield(ta, :timestamp)
 _values(ta::TimeArray)    = getfield(ta, :values)
 _colnames(ta::TimeArray)  = getfield(ta, :colnames)
-_meta(ta::TimeArray)	  = getfield(ta, :meta)
+_meta(ta::TimeArray)      = getfield(ta, :meta)
```
ararslan: Why are these definitions needed in addition to the others?

iblis17: well, internal use like this case:
https://github.com/JuliaStats/TimeSeries.jl/blob/master/src/combine.jl#L14

There are a keyword argument named as `colnames` already,
so the getter function is unavailable in that function.

ararslan: I see, okay.

PR #390, opened 2018-10-30
improve rename

### LINKED ISSUE #387: Rename column names using a Dict{Symbol, Symbol} or Pair{Symbol, Symbol} or several Pair(s)
Hello,

TimeSeries have a `rename` function http://juliastats.github.io/TimeSeries.jl/dev/modify/#rename-1
https://github.com/JuliaStats/TimeSeries.jl/blob/3fe3490d9802b559422e810d84eced8e35c25a16/src/modify.jl#L35-L42

but implementing a `rename` column names function using a `Dict` could help https://github.com/JuliaQuant/MarketData.jl/issues/57

Possible usage

```
d = Dict(
    :Close=> :Close2
)
rename(ts, d)
```

Possible implementation

```
    _colnames = copy(colnames(cl))
    for (i, colname) in enumerate(_colnames)
        if colname in keys(d_ren)
            _colnames[i] = d_ren[colname]
        end
    end
    rename(ta, _colnames)
```

Similar to https://github.com/dysonance/Temporal.jl/issues/22

Kind regards

### LINKED ISSUE #389: Rename column names using a Function Symbol->Symbol
Similar to https://github.com/JuliaStats/TimeSeries.jl/issues/387

Implementing

```julia
rename(f::Function, ta::TimeArray)
```

might be considered.

could help https://github.com/JuliaQuant/MarketData.jl/issues/57

femtotrader:
Closes #387
Closes #389


src/modify.jl
```diff
@@ -40,3 +40,26 @@ function rename(ta::TimeArray, colnames::Vector{Symbol})
...
+        if colname in keys(d_colnames)
+            _colnames[i] = d_colnames[colname]
+        end
+    end
+    rename(ta, _colnames)
+end
+
+rename(ta::TimeArray, p_colnames::Pair{Symbol, Symbol}) = rename(ta, Dict(p_colnames))
+
+function rename(f::Function, ta::TimeArray; asstring=false)
```
iblis17: I think the `f::Base.Callable` (which is new stuff in v0.7/1.0) is better

src/modify.jl
```diff
@@ -40,3 +40,26 @@ function rename(ta::TimeArray, colnames::Vector{Symbol})
 end
 
 rename(ta::TimeArray, colnames::Symbol) = rename(ta, [colnames])
+
+function rename(ta::TimeArray, d_colnames::Dict{Symbol, Symbol})
```
iblis17: I think we need to raise error, if there are duplicated values in the dictionary.
e.g. If user input this
```
Dict(:Open => :a, :Close => :a)
```

iblis17: hmm, I forgot that we will automatically change the duplicated column names.
Ignore this.

docs/src/modify.md
```diff
@@ -31,4 +31,8 @@ using MarketData
 rename(cl, :Close′)
 rename(cl, [:Close′])
 rename(ohlc, [:Open′, :High′, :Low′, :Close′])
+rename(ohlc, :Open => :Open′)
+rename(ohlc, Dict(:Open => :Open′, :Close => :Close′))
+rename(Symbol ∘ uppercase ∘ string, ohlc)
+rename(uppercase, ohlc, asstring=true)
```
femtotrader: I don't think this is a great API idea maybe we should use type dispatch instead and use

    rename(uppercase, ohlc, String)

src/modify.jl
femtotrader: Absolutely

```julia
julia> rename(ohlc, [:Open, :Open, :Low, :Close])
500×4 TimeArray{Float64,2,Date,Array{Float64,2}} 2000-01-03 to 2001-12-31
│            │ Open   │ Open_1 │ Low    │ Close  │
├────────────┼────────┼────────┼────────┼────────┤
│ 2000-01-03 │ 104.88 │ 112.5  │ 101.69 │ 111.94 │
```

even without duplicated values it could lead to duplicated coumn names and so this is fixed at an other layer.

```julia
julia> rename(ohlc, Dict(:High=>:Open))
500×4 TimeArray{Float64,2,Date,Array{Float64,2}} 2000-01-03 to 2001-12-31
│            │ Open   │ Open_1 │ Low    │ Close  │
├────────────┼────────┼────────┼────────┼────────┤
│ 2000-01-03 │ 104.88 │ 112.5  │ 101.69 │ 111.94 │
```

src/modify.jl
```diff
@@ -53,13 +53,19 @@ end
...
     end
     rename(ta, _colnames)
 end
+
+function rename(f::Base.Callable, ta::TimeArray, colnametyp::Type{String})
+    f = Symbol ∘ f ∘ string
+    rename(f, ta)
+end
+
+function rename(f::Base.Callable, ta::TimeArray)
```
iblis17: I think this function is same as
`function rename(f::Base.Callable, ta::TimeArray, colnametyp::Type{Symbol} = Symbol)`

docs/src/modify.md
```diff
@@ -31,4 +31,8 @@ using MarketData
 rename(cl, :Close′)
 rename(cl, [:Close′])
 rename(ohlc, [:Open′, :High′, :Low′, :Close′])
+rename(ohlc, :Open => :Open′)
+rename(ohlc, Dict(:Open => :Open′, :Close => :Close′))
```
femtotrader: An other approach could be to support multiple pairs for replacement

    rename(ohlc, :Open => :Open′, :Close => :Close′)

or

    rename(ohlc, Dict(:Open => :Open′, :Close => :Close′)...)

it seems to be where `replace` function should be going according

`https://github.com/JuliaLang/julia/issues/29849#issuecomment-434321585`

iblis17: + 1 for `Vararg` pairs, so we do not need to support Dict explicitly,
just show the usage in doc is enough.

femtotrader: Absolutely so we could do

    rename(ohlc, Dict(:Open => :Open′, :Close => :Close′)...)

which is enough

src/modify.jl
```diff
@@ -40,3 +40,27 @@ function rename(ta::TimeArray, colnames::Vector{Symbol})
 end
 
 rename(ta::TimeArray, colnames::Symbol) = rename(ta, [colnames])
+
+function rename(ta::TimeArray, args...)
```
iblis17: IIRC, this is legit:
`function rename(ta::TimeArray, args::Pair{Symbol,Symbol}...)`

PR #391, opened 2018-10-30
docs: constructors of TimeArray

iblis17:


docs/src/timearray.md
```diff
@@ -70,3 +70,9 @@ values
 colnames
 meta
 ```
+
+## Constructors
```
femtotrader: I wonder if doc for TimeArray constructors shouldn't be written before "Fields getter functions"

Maybe other functions which have a TimeArray as parameter should all be listed here (or in a separated page)

like http://juliadata.github.io/DataFrames.jl/latest/lib/functions.html

docs/Makefile
```diff
@@ -1,2 +1,2 @@
 all:
-	julia --color=yes make.jl
+	julia --project=. --color=yes make.jl
```
femtotrader: I'd be very pleased to know why it's necessary (for my own information)

docs/src/timearray.md
iblis17: > I wonder if doc for TimeArray constructors shouldn't be written before "Fields getter functions"

okay, I can reagrange the sections.

>Maybe other functions which have a TimeArray as parameter should all be listed here (or in a separated page)

I think you want an API referece list, this is another issue.

docs/Makefile
iblis17: Since we have `Project.toml` in that dir.

`julia --project=.` implies `julia -e 'using Pkg; Pkg.activate(".")'`

iblis17: And you can checkout the doc deployment instructions in `.travis.yml` as well

PR #395, opened 2018-11-23
Construct TimeArray with NamedTuple

### LINKED ISSUE #394: Construct TimeArray from NamedTuple
Hello,

DataFrame from [DataFrames.jl](https://github.com/JuliaData/DataFrames.jl) can be construct from `NamedTuple`

```julia
julia> using Dates

julia> using DataFrames

julia> data = (time=[DateTime(2018, 11, 21, 12, 0), DateTime(2018, 11, 21, 13, 0)], col1=[10.2, 11.2], col2=[20.2, 21.2])
(time = DateTime[2018-11-21T12:00:00, 2018-11-21T13:00:00], col1 = [10.2, 11.2], col2 = [20.2, 21.2])

julia> DataFrame(data)
2×3 DataFrame
│ Row │ time                │ col1    │ col2    │
│     │ DateTime            │ Float64 │ Float64 │
├─────┼─────────────────────┼─────────┼─────────┤
│ 1   │ 2018-11-21T12:00:00 │ 10.2    │ 20.2    │
│ 2   │ 2018-11-21T13:00:00 │ 11.2    │ 21.2    │
```

It will be nice if `TimeArray` could also be constructed like this.

```
julia> using TimeSeries

julia> TimeArray(data)
```

maybe passing name of datetime column should be considered (as a `Symbol`).

```julia
julia> TimeArray(data, timestamp=:time)
```

LibPQ uses this kind of mechanism
https://invenia.github.io/LibPQ.jl/latest/index.html

Kind regards

PS : similar to https://github.com/dysonance/Temporal.jl/issues/33

scls19fr:
Closes #394

### DIFF
--- a/src/timearray.jl
+++ b/src/timearray.jl
@@ -13,8 +13,7 @@ abstract type AbstractTimeSeries{T,N,D} end
 
     TimeArray(timestamp, values[, colnames, meta=nothing])
     TimeArray(ta::TimeArray; timestamp, values, colnames, meta)
-    data = (datetime=[DateTime(2018, 11, 21, 12, 0), DateTime(2018, 11, 21, 13, 0)], col1=[10.2, 11.2], col2=[20.2, 21.2], col3=[30.2, 31.2])
-    ta = TimeArray(data; timestamp=:datetime, meta="Example")
+    TimeArray(data::NamedTuple; timestamp, meta)
 
 The second constructor will yields a new TimeArray with the new given fields.
 Note that the unchanged fields will be shared, there aren't any copy for the
@@ -34,6 +33,12 @@ The third constructor build a TimeArray from a NamedTuple.
   the column of `values`.
 
 - `meta::Any`: a user-defined metadata.
+
+# Examples
+
+    data = (datetime=[DateTime(2018, 11, 21, 12, 0), DateTime(2018, 11, 21, 13, 0)], col1=[10.2, 11.2], col2=[20.2, 21.2], col3=[30.2, 31.2])
+    ta = TimeArray(data; timestamp=:datetime, meta="Example")
+
 """
 struct TimeArray{T,N,D<:TimeType,A<:AbstractArray{T,N}} <: AbstractTimeSeries{T,N,D}
 
--- a/src/timearray.jl
+++ b/src/timearray.jl
@@ -80,6 +80,12 @@ TimeArray(ta::TimeArray;
           colnames = _colnames(ta), meta = _meta(ta), args...) =
     TimeArray(timestamp, values, colnames, meta; args...)
 
+function TimeArray(data::NamedTuple; timestamp=:time)
+    columns = (key for key in keys(data) if key != timestamp)
+    dat = hcat((data[key] for key in columns)...)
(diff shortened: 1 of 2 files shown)


scls19fr:
Missing passing `meta` and improving docstring

src/timearray.jl
```diff
@@ -13,11 +13,15 @@ abstract type AbstractTimeSeries{T,N,D} end
 
     TimeArray(timestamp, values[, colnames, meta=nothing])
     TimeArray(ta::TimeArray; timestamp, values, colnames, meta)
+    data = (datetime=[DateTime(2018, 11, 21, 12, 0), DateTime(2018, 11, 21, 13, 0)], col1=[10.2, 11.2], col2=[20.2, 21.2], col3=[30.2, 31.2])
+    ta = TimeArray(data; timestamp=:datetime, meta="Example")
```
iblis17: I will just put the function prototype here.
`TimeArray(data::NamedTuple; timestamp, meta)`

and add a `Examples` section, like the docstring in Base
```
# Examples
date = ...
ta = ...
```

src/timearray.jl
```diff
@@ -80,6 +84,12 @@ TimeArray(ta::TimeArray;
           colnames = _colnames(ta), meta = _meta(ta), args...) =
     TimeArray(timestamp, values, colnames, meta; args...)
 
+function TimeArray(data::NamedTuple; timestamp=:time, meta=nothing, args...)
```
iblis17: I think we do not need default value for `timestamp`.

scls19fr: `timestamp` here is the name of the field of `NamedTuple` that store datetime... we must define a default value

iblis17: yeah, that field indicate the time index.
I think a required keyword argument is better, not optional, then user need to set it explicitly.

src/timearray.jl
```diff
@@ -89,7 +89,7 @@ TimeArray(ta::TimeArray;
           colnames = _colnames(ta), meta = _meta(ta), args...) =
     TimeArray(timestamp, values, colnames, meta; args...)
 
-function TimeArray(data::NamedTuple; timestamp=:time, meta=nothing, args...)
+function TimeArray(data::NamedTuple, timestamp::Symbol; meta=nothing, args...)
```
iblis17: I mean required keyword argument:
```julia
TimeArray(data::NamedTuple; timestamp, meta = nothing, args...)
```

test/timearray.jl
```diff
@@ -189,7 +189,7 @@ end
 
 @testset "construct with NamedTuple" begin
     data = (datetime=[DateTime(2018, 11, 21, 12, 0), DateTime(2018, 11, 21, 13, 0)], col1=[10.2, 11.2], col2=[20.2, 21.2], col3=[30.2, 31.2])
-    ta = TimeArray(data, :datetime; meta="Example")
+    ta = TimeArray(data; timestamp=:datetime, meta="Example")
```
iblis17: well, you add the default value back.
I want no default value: https://github.com/JuliaStats/TimeSeries.jl/pull/395#discussion_r236046606

iblis17:
Thanks! :+1:

PR #399, opened 2019-02-23
Revise show method for timearray

jjstickel:
Now a one-line short version will be output (when no MIME type is specified). The previous long version will be output for ::MIME"text/plain", e.g., when printing in the REPL. The tests were also updated.

The motivation for this change comes from including timearray objects in structs and dicts. Printing of the parent object resulted in the full printing of the timearray instead of a summary statement. This PR enables the printing of just a summary statement in such a case.

### DIFF
--- a/src/timearray.jl
+++ b/src/timearray.jl
@@ -208,7 +208,7 @@ calculate the paging
     ret
 end
 
-function show(io::IO, ta::TimeArray{T}) where T
+function print_time_array(io::IO, ta::TimeArray{T}, short=false) where T
     # summary line
     nrow = size(values(ta), 1)
     ncol = size(values(ta), 2)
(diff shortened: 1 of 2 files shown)


iblis17:
```julia
julia> [cl, cl]
2-element Array{TimeArray{Float64,1,Date,Array{Float64,1}},1}:
 500×1 TimeArray{Float64,1,Date,Array{Float64,1}} 2000-01-03 to 2001-12-31
 500×1 TimeArray{Float64,1,Date,Array{Float64,1}} 2000-01-03 to 2001-12-31

```

jjstickel:
This has been approved, but not merged. Anything I need to do? I don't have permissions to merge myself.

iblis17:
@jjstickel Thanks for your great contributions! :+1:

jd-lara:
@iblis17 can we get a tag to get this PR into our code base?

iblis17:
ah, sure.

iblis17:
xref: https://github.com/JuliaLang/METADATA.jl/pull/22774

PR #403, opened 2019-04-28
Insert an explicit call example for vcat().

petershintech:


docs/src/combine.md
```diff
@@ -107,7 +107,8 @@ For example:
 using TimeSeries
 a = TimeArray([Date(2015, 10, 01), Date(2015, 11, 01)], [15, 16])
 b = TimeArray([Date(2015, 12, 01)], [17])
-[a; b]
+vcat(a,b)
```
iblis17:
```suggestion
vcat(a, b)
```

PR #405, opened 2019-04-30
Be specific about exception type.

petershintech:
Raise a more specific ArgumentError when an unsupported method is given to dropnan().


petershintech:
And also for percentage() and setcolnames().

petershintech:
@iblis17 I found bdee40c is breaking some unit tests. Is it ok to merge it to the master?

petershintech:
Thanks @iblis17

PR #407, opened 2019-05-03
Plotting: labels as strings

greimel:
This is a workaround for a bug in Plots.pgfplots() which prevents labels::Symbol to be shown.

see JuliaPlots/Plots.jl#2020

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -7,7 +7,7 @@
     #    ta, ohlc = extract_ohlc(ta)
     #    collect(zip(ohlc)) # But there are currently issues with that
     else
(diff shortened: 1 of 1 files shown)


greimel:
```julia
using Dates, TimeSeries
# dummy data
dates = Date(2018, 1, 1):Day(1):Date(2018, 12, 31)
ta = TimeArray(dates, rand(length(dates)))

using Plots
gr()
plot(ta)
savefig("test-gr.png")
pgfplots()
plot(ta)
savefig("test-pgf-before.png")
```


before:
test-pgf-before

after:
test-pgf

cf gr():
test-gr

PR #433, opened 2019-10-09
Replace find -> findall and fix issue 432

### LINKED ISSUE #432: Plotting TimeArray with seriestype == :candlestick causes exception "UndefVarError: find not defined"
When plotting a `TimeArray` using `seriestype = :candlestick`, I get the following exception:

```
UndefVarError: find not defined
```

caused at [plotrecipes.jl:27](https://github.com/JuliaStats/TimeSeries.jl/blob/v0.16.0/src/plotrecipes.jl#L27). I'm using v0.16.0 of TimeSeries.jl and Julia 1.12. Indeed, the function `find` is not defined. The code in `plotrecipes.jl` seems not to have changed between v0.16.0 and `master`.

Sh4pe:
The function find is not defined and caused an UndefVarError
when this line was executed.

Fixes #432

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,11 @@
+### 0.16.1
+
+* Fix Candlestick plotting. ([#433])
+
+
+[#433]: https://github.com/JuliaStats/TimeSeries.jl/pull/433
+
+
 ### 0.16.0
 
 * Improve performance of `moving` function. (#414)
--- a/docs/src/plotting.md
+++ b/docs/src/plotting.md
@@ -34,7 +34,8 @@ found in the Plots
 [documentation](http://docs.juliaplots.org/latest/supported/).
 
 
-Plotting candlestick:
+Plotting candlestick, the ``Candlestick`` constructor requires four columns
+exist in the input. They are `open`, `high`, `low` and `close` (case-insensitive).:
 
 ```julia
 plot(TimeSeries.Candlestick(MarketData.ohlc))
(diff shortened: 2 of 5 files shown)


iblis17:
Hi @Sh4pe could you test the updated patches?

Sh4pe:
It still does not work and fails inside `extract_ohlc`. I've added a test that invokes the `Candlestick` constructor and [it fails](https://travis-ci.org/JuliaStats/TimeSeries.jl/jobs/598303378#L375) with the error message 

```
ArgumentError: invalid index: nothing of type Nothing
```
(The [Julia 1.0 build job](https://travis-ci.org/JuliaStats/TimeSeries.jl/jobs/598303379#L284) failed because it could not connect to GitHub. Maybe it needs to be rekicked).

iblis17:
ah, I fixed the error message.
A Candlestick requires theses four columns as inputs: `open`, `high`, `low`, `close`.
A worked example:
```julia
using MarketData
TimeSeries.Candlestick(ohlcv)
```

Sh4pe:
Ok, got it. Didn't know that these columns have to be present. Maybe a short comment in [plotting.md](https://github.com/JuliaStats/TimeSeries.jl/blob/master/docs/src/plotting.md) should be added saying that these columns are required for a candlestick plot. What do you think?

iblis17:
Thanks for your contribution.

I'm going to make a new release.

PR #440, opened 2019-12-11
Added maximum and minimum from Base as reductions

ohmsweetohm1:


iblis17:
Thanks for your contributions. :+1

PR #445, opened 2020-03-23
recipesbase compat

mkborregaard:
Should be 100% safe. Might we get a release with this please? This is one of the packages holding back upgrading RecipesBase (and by extension Plots).


iblis17:
Ah sure, I will tag a new release

PR #453, opened 2020-05-07
Fix warning

IljaK91:


iblis17:
Hi, are there some references about this change?

IljaK91:
I do not have any links. I just tried to plot a TimeArray and received a warning, that "label" should be used instead of "labels". I fixed it locally and it stopped showing warnings, so I guess it was the right thing to fix.

iblis17:
ah, okay.

PR #489, opened 2021-03-17
Modified show, added omitted columns message

### LINKED ISSUE #468: Add omitting of columns when printing
When printing a TimeArray in the REPL even for a TimeArray with 1000 columns all of them are printed.

rafaelchp:
Hello, this is a proposal to fix #468. It prints page 1 of `TimeArray` and adds the omitted columns message. I would like some feedback on it. 

I have not rewritten the tests yet, but I'd gladfully do it.


iblis17:
I just found that `DataFrame` puts the message in the right hand side and change the color. Maybe we can do that too?

src/timearray.jl
```diff
@@ -259,60 +261,89 @@ function print_time_array(io::IO, ta::TimeArray{T}, short=false) where T
...
-                for j in p
-                    print(io, "│ ", _rpad(strs[i, j], colwidth[j] + 1))
-                end
-                print(io, "│")
-            end
+    _print_page(io, pages[1], ta, 
+        spacetime, colwidth, nrow, drow, res_row, tophalf, bothalf, 
+        length(pages))
+    
+    if length(pages) > 1
```
iblis17, line 268: I think we need to support `@show` for printing all pages out.

rafaelchp: Yes, that would be great and it would take advantage of the paging machinery. I'm not sure how to do this, though; I tried looking at the `DataFrames` package and to `@show df` with a large `DataFrame` and it dumps all the data on screen (not readable) and then the same trimmed view of the `df`. I am not sure of what methods to override. Or maybe a separate method for this?

iblis17: Oh, I guess `DataFrames` use the `:limit` option of  `IOContext` to control the omitting.

You can try this:
```
show(IOContext(stdout, :limit => false), df)
```

vs 

```
show(df)
```

I think we can also utilize that option.

src/timearray.jl
```diff
@@ -259,63 +261,111 @@ function print_time_array(io::IO, ta::TimeArray{T}, short=false) where T
...
+    end
 end
-Base.show(io::IO, ta::TimeArray) = print_time_array(io, ta, true)
-Base.show(io::IO, ::MIME"text/plain", ta::TimeArray) =
-    print_time_array(io, ta, false)
+
+Base.show(io::IO, ta::TimeArray) = 
+    print_time_array(io, ta, get(io, :limit, false), get(io, :limit, true))
+Base.show(io::IO, ::MIME"text/plain", ta::TimeArray) = 
+    print_time_array(io, ta, false, !get(io, :limit, false))
```
rafaelchp, line 366: I checked the `:limit` in the `show` methods definition and added a new parameter in `print_time_array`

rafaelchp: It is working as follows: 
- Showing a `TimeArray` in the REPL by just calling it by name (say `ta`) prints omitted columns message if the `TimeArray` has lots of columns.
- Calling `@show ta` prints all columns and then the compact view. Similar to `DataFrames` but the first output a little more ordered.

I had to modify some tests to pass the `:limit` option of `IOContext`

src/timearray.jl
```diff
@@ -259,63 +261,111 @@ function print_time_array(io::IO, ta::TimeArray{T}, short=false) where T
...
+Base.show(ta::TimeArray) =
```
iblis17, line 367: I think we don't need this line?

src/timearray.jl
iblis17: I'm investigating why `show(ta)` and `@show ta` look different. :thinking:

src/timearray.jl
```diff
@@ -259,63 +261,111 @@ function print_time_array(io::IO, ta::TimeArray{T}, short=false) where T
...
         end
-    end  # for p ∈ pages
+    end
 end
-Base.show(io::IO, ta::TimeArray) = print_time_array(io, ta, true)
-Base.show(io::IO, ::MIME"text/plain", ta::TimeArray) =
-    print_time_array(io, ta, false)
+
+Base.show(io::IO, ta::TimeArray) = 
+    print_time_array(io, ta, get(io, :limit, false), get(io, :limit, true))
```
iblis17, line 364: I found that `summary` is a more proper method for the setting `short=true`.

So how about this?
```suggestion
Base.summary(io::IO, ta::TimeArray) = print_time_array(io, ta, true)

Base.show(io::IO, ta::TimeArray) = 
    print_time_array(io, ta, false, get(io, :limit, true))
```

test/timearray.jl
```diff
@@ -497,21 +497,20 @@ end
...
 │ 2001-12-28 │ 22.43  │
-│ 2001-12-31 │ 21.9   │"""
+│ 2001-12-31 │ 21.9   │
+"""
         @test str == out
     end
 
     # my edits above seem to work -- now need to do for the rest, JJS 2/22/19
-    let str = sprint(show, ohlc)
+    let str = sprint(show, ohlc; context = :limit=>true)
```
iblis17, line 513: Then, this case is just the `Base.summary`
```suggestion
    let str = sprint(summary, ohlc)
```

src/timearray.jl
iblis17: ah, I figured it out, our implementation is correct. 
`@show` invoking `sprint` with `IOBuffer` for rendering, so the return of `displaysize(io)` is different from `displaysize(stdout)`.

test/timearray.jl
```diff
@@ -528,44 +527,65 @@ end
...
+│ ⋮          │ ⋮      │ ⋮      │ ⋮      │ ⋮      │
 │ 2001-12-20 │ 21.4   │ 21.47  │ 20.62  │ 20.67  │
 │ 2001-12-21 │ 21.01  │ 21.54  │ 20.8   │ 21.0   │
 │ 2001-12-24 │ 20.9   │ 21.45  │ 20.9   │ 21.36  │
 │ 2001-12-26 │ 21.35  │ 22.3   │ 21.14  │ 21.49  │
 │ 2001-12-27 │ 21.58  │ 22.25  │ 21.58  │ 22.07  │
 │ 2001-12-28 │ 21.97  │ 23.0   │ 21.96  │ 22.43  │
-│ 2001-12-31 │ 22.51  │ 22.66  │ 21.83  │ 21.9   │"""
+│ 2001-12-31 │ 22.51  │ 22.66  │ 21.83  │ 21.9   │
+"""
```
iblis17, line 538: Is the newline added intended?

I prefer the output without newline, since the REPL will add one while rendering.

src/timearray.jl
rafaelchp: Yeah, not really, I thought DataFrames used this to show the complete DataFrame to the stdout. I will remove it.

test/timearray.jl
rafaelchp: You're right, I think it works better with `summary`. DataFrames do this too. I'll review the test suite with some changes on this front and let's see how it goes.

test/timearray.jl
rafaelchp: You're right. I think I had it to separate the full printing of the TimeArray from the one with the omitted columns when `@show`ing.

test/timearray.jl
```diff
@@ -480,7 +480,7 @@ end
 
 @testset "show methods don't throw errors" begin
     io = IOBuffer()
-    let str = sprint(show, cl)
+    let str = sprint(summary, cl)
         out = "500×1 TimeArray{Float64,1,Date,Array{Float64,1}} 2000-01-03 to 2001-12-31"
```
rafaelchp, line 484: It seems that tests are failing in 1.6 because `Array{Float64,1}` is now shown as `Vector{Float64}` and `Array{Float64,2}` as `Matrix{Float64}`.

rafaelchp: Maybe in `test/timearray.jl` we can add a string to interpolate in the output depending on the value of `VERSION`. Something like: 

```julia
dispvec = VERSION < v"1.6" ? "Array{Float64,1}" : "Vector{Float64}"
```

Or maybe even typed: 
```julia
dispvec(T) = VERSION < v"1.6" ? "Array{$(repr(T)),1}" : "Vector{$(repr(T))}"
```

iblis17: ah, fine. :+1:

rafaelchp: Turns out that this `TimeArray{Float64,1,Date,Array{Float64,1}}` becomes `TimeArray{Float64, 1, Date, Vector{Float64}}` in 1.6.

iblis17:
Thanks for your contributions. :+1:

rafaelchp:
Glad to help 😬

PR #493, opened 2021-04-16
Fix plots recipes

### LINKED ISSUE #492: Plot warnings in Documentation
In https://juliastats.org/TimeSeries.jl/stable/plotting/ the plots don't have the correct colors anymore and
it shows the following warning:

```
┌ Warning: Indices Base.OneTo(24) of attribute `seriescolor` does not match data indices 1:263.
└ @ Plots ~/.julia/packages/Plots/SVksJ/src/utils.jl:102
┌ Info: Data contains NaNs or missing values, and indices of `seriescolor` vector do not match data indices.
│ If you intend elements of `seriescolor` to apply to individual NaN-separated segements in the data,
│ pass each segment in a separate vector instead, and use a row vector for `seriescolor`. Legend entries
│ may be suppressed by passing an empty label.
│ For example,
└     plot([1:2,1:3], [[4,5],[3,4,5]], label=["y" ""], seriescolor=[1 2])
```

I suspect that something on the Plots.jl side changed, since it also poped up suddenly in my own code.

iblis17:
fix #492

### DIFF
--- a/src/plotrecipes.jl
+++ b/src/plotrecipes.jl
@@ -83,7 +83,7 @@ end
         throw(ArgumentError(":seriescolor should be a Vector{Symbol} of two elements."))
     end
 
-    xseg, yseg = Float64[], Float64[]
+    xseg, yseg = Vector{Float64}[], Vector{Float64}[]
     xcenter, ycenter = Float64[], Float64[]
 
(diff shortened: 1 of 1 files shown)


iblis17:
I'm going to merge this tmr if no objection.

PR #496, opened 2021-06-13
Add intersect method

guilhermebodin:


src/timeaxis/timegrid.jl
```diff
@@ -58,21 +58,34 @@ Base.eltype(::Type{<:TimeGrid{T}}) where T = T
...
+end
 
 function Base.show(io::IO, ::MIME{Symbol("text/plain")}, tg::TimeGrid{T,P,:finite}) where {T,P}
     summary(io, tg)
-    println(io, ":")
-    print(io, " $(tg.o)\n",
-              "  ⋮\n",
-              " $(tg[end])\n",
-              " / $(tg.p)")
+    if !isempty(tg)
```
iblis17, line 82: How about the early return?
```suggestion
    isempty(tg) && return
```

src/timeaxis/timegrid.jl
```diff
@@ -58,21 +58,34 @@ Base.eltype(::Type{<:TimeGrid{T}}) where T = T
...
+Base.isfinite(tg::TimeGrid{T,P,:infinite}) where {T,P} = false
 
 ###############################################################################
 #  Printing
 ###############################################################################
 
-Base.show(io::IO, tg::TimeGrid{T,P,:finite}) where {T,P} =
-    print(io, "$(tg.o) … $(tg[end]) / $(tg.p) ")
+function Base.show(io::IO, tg::TimeGrid{T,P,:finite}) where {T,P}
+    summary(io, tg)
```
iblis17, line 74: well, I won't invoke `summary` here, since this function will be used for `repr`.
```suggestion

```

src/timeaxis/timegrid.jl
```diff
@@ -287,6 +300,45 @@ Base.resize!(tg::TimeGrid{T,P,:infinite}, n::Int) where {T,P} = tg
...
+    for tg in tgs
+        tg1 = intersect(tg1, tg)
+    end
+    return tg1
+end
+# TODO 
+# Consider checking if they are equal
+function intersect(tg_inf::TimeGrid{T,P,:infinite}, 
+                   tg_fin::TimeGrid{T,P,:finite})::TimeGrid{T,P,:finite} where {T,P}
+    @assert tg_fin.p == tg_inf.p
```
iblis17, line 316: Could we also have the support for where `tg_fin.p / tg_inf.p` is a rational number?

guilhermebodin: I think we can but I suggest we open an issue and leave for another PR

src/timeaxis/timegrid.jl
```diff
@@ -58,21 +58,33 @@ Base.eltype(::Type{<:TimeGrid{T}}) where T = T
...
 ###############################################################################
 #  Printing
 ###############################################################################
 
-Base.show(io::IO, tg::TimeGrid{T,P,:finite}) where {T,P} =
+function Base.show(io::IO, tg::TimeGrid{T,P,:finite}) where {T,P}
+    isempty(tg) && return
     print(io, "$(tg.o) … $(tg[end]) / $(tg.p) ")
+    return
+end
```
iblis17, line 77: hmm, something like this for empty case?
```suggestion
Base.show(io::IO, tg::TimeGrid{T,P,:finite}) where {T,P} = 
    isempty(tg) ? print(io, "(empty)") : print(io, "$(tg.o) … $(tg[end]) / $(tg.p)")
```

src/timeaxis/timegrid.jl
```diff
@@ -58,21 +58,33 @@ Base.eltype(::Type{<:TimeGrid{T}}) where T = T
...
+    isempty(tg) && return
     println(io, ":")
     print(io, " $(tg.o)\n",
-              "  ⋮\n",
-              " $(tg[end])\n",
-              " / $(tg.p)")
+            "  ⋮\n",
+            " $(tg[end])\n",
+            " / $(tg.p)")
+    return
```
iblis17, line 87: minor style changes
```suggestion
              "  ⋮\n",
              " $(tg[end])\n",
              " / $(tg.p)")
```

src/timeaxis/timegrid.jl
```diff
@@ -58,21 +58,33 @@ Base.eltype(::Type{<:TimeGrid{T}}) where T = T
 Base.length(tg::TimeGrid{T,P,:finite}) where {T,P} = tg.n
 Base.size(tg::TimeGrid{T,P,:finite}) where{T,P}    = tg.n
 
+###############################################################################
+#  Empty TimeGrid
+###############################################################################
+Base.isempty(tg::TimeGrid{T,P,:finite}) where {T,P} = iszero(tg.n)
+Base.isempty(tg::TimeGrid{T,P,:infinite}) where {T,P} = false
+Base.isfinite(tg::TimeGrid{T,P,:finite}) where {T,P} = true
+Base.isfinite(tg::TimeGrid{T,P,:infinite}) where {T,P} = false
```
iblis17, line 67: minor style changes
```suggestion

Base.isempty(tg::TimeGrid{T,P,:finite}) where {T,P}   = iszero(tg.n)
Base.isempty(tg::TimeGrid{T,P,:infinite}) where {T,P} = false

Base.isfinite(tg::TimeGrid{T,P,:finite}) where {T,P}   = true
Base.isfinite(tg::TimeGrid{T,P,:infinite}) where {T,P} = false
```

src/timeaxis/timegrid.jl
```diff
@@ -287,6 +299,45 @@ Base.resize!(tg::TimeGrid{T,P,:infinite}, n::Int) where {T,P} = tg
...
+intersect(tg_fin::TimeGrid{T,P,:finite}, tg_inf::TimeGrid{T,P,:infinite}) where {T,P} = intersect(tg_inf, tg_fin)
+function intersect(tg_inf1::TimeGrid{T,P,:infinite}, tg_inf2::TimeGrid{T,P,:infinite}) where {T,P}
+    @assert tg_inf1.p == tg_inf2.p
+    new_o = max(tg_inf1.o, tg_inf2.o)
+    return TimeGrid(new_o, tg_inf1.p)
+end
+function intersect(tg_fin1::TimeGrid{T,P,:finite}, tg_fin2::TimeGrid{T,P,:finite}) where {T,P}
+    @assert tg_fin1.p == tg_fin2.p
+    if (tg_fin2[end] < tg_fin1.o) || (tg_fin1[end] < tg_fin2.o)
+        return TimeGrid(tg_fin1.o, tg_fin1.p, 0)
```
iblis17, line 333: There is a constructor that can help for this case
```suggestion
        return TimeGrid(tg_fin1, n = 0)
```

src/timeaxis/timegrid.jl
```diff
@@ -287,6 +299,45 @@ Base.resize!(tg::TimeGrid{T,P,:infinite}, n::Int) where {T,P} = tg
...
+end
+# TODO 
+# Consider checking if they are equal
+function intersect(tg_inf::TimeGrid{T,P,:infinite}, 
+                   tg_fin::TimeGrid{T,P,:finite})::TimeGrid{T,P,:finite} where {T,P}
+    @assert tg_fin.p % tg_inf.p
+    if tg_fin[end] < tg_inf.o
+        return TimeGrid(tg_fin.o, tg_fin.p, 0)
+    end
+    new_o = max(tg_fin.o, tg_inf.o)
```
iblis17, line 319: Is it okay with you, if I rename `new_o` to `o′` someday?
Personally, I prefer shorter name. :p

guilhermebodin: no problem

src/timeaxis/timegrid.jl
```diff
@@ -287,6 +299,45 @@ Base.resize!(tg::TimeGrid{T,P,:infinite}, n::Int) where {T,P} = tg
...
+                                  findfirst(isequal(tg_fin[end]), tg_fin)
+    return TimeGrid(new_o, tg_fin.p, new_n)
+end
+intersect(tg_fin::TimeGrid{T,P,:finite}, tg_inf::TimeGrid{T,P,:infinite}) where {T,P} = intersect(tg_inf, tg_fin)
+function intersect(tg_inf1::TimeGrid{T,P,:infinite}, tg_inf2::TimeGrid{T,P,:infinite}) where {T,P}
+    @assert tg_inf1.p == tg_inf2.p
+    new_o = max(tg_inf1.o, tg_inf2.o)
+    return TimeGrid(new_o, tg_inf1.p)
+end
+function intersect(tg_fin1::TimeGrid{T,P,:finite}, tg_fin2::TimeGrid{T,P,:finite}) where {T,P}
```
iblis17, line 330: I found a case that might yield empty TimeGrid:
```julia
julia> tg = TimeGrid(DateTime(2021, 1, 1, 0), Hour(1), 24)
24-element TimeGrid{DateTime, Hour, :finite}:
 2021-01-01T00:00:00
  ⋮
 2021-01-01T23:00:00
 / 1 hour

julia> tg′ = TimeGrid(DateTime(2021, 1, 1, 0, 30), Hour(1), 24)
24-element TimeGrid{DateTime, Hour, :finite}:
 2021-01-01T00:30:00
  ⋮
 2021-01-01T23:30:00
 / 1 hour

julia> intersect(collect(tg), collect(tg′))
DateTime[]

```

guilhermebodin: In fact I was wrong on slack there is a way to infinite with infinite intersection being empty

guilhermebodin: I rewrote some parts and added more tests. 

The current behavior of intersecting tg with different periods is method error. What do you think about that?

iblis17: > The current behavior of intersecting tg with different periods is method error. What do you think about that?

That might need `IrregularTimeGrid` implemented.
I think we can revisit it later, putting a `TODO` here is fine for me.

I still don't make a perfect algo come out. Here is my two cents:
- For the case `tg.p <: Period` (or maybe narrow down to `FixedPeriod` ?), treat all the period as `Int` and we can get a `TimeGrid`.
- For the case `tg.p <: Float64` (and `Irrational` included. Although I know `π` is marked a `Irrational` type, all calculation on it are approximations.), just fallback to the `intersect(::Vector, ::Vector)` calculation and output `IrregularTimeGrid`.

PR #503, opened 2021-12-07
Clean up tables interface implementation

quinnj:
Reasoning for changes:
  * We don't need to define `rowaccess`/`rows` because if we define `columnaccess`/`columns` then we automatically get an efficient fallback `rows` definition
  * We don't need to define `columnnames`/`getcolumn`/`schema` on `TimeArray`, only on the type that `Tables.columns` _returns_, so in this case, just `TableIter`
  * In the constructor, we need to call `Tables.columns` first and _then_ call `Tables.schema(cols)`; calling `Tables.schema` directly on the table input isn't valid, as it wouldn't work in, for example, the case where someone passed in a `::TimeArray`.

### DIFF
--- a/src/tables.jl
+++ b/src/tables.jl
@@ -48,17 +48,11 @@ function Base.iterate(x::TableIter, i::Integer = 1)
 end
 
 Tables.istable(::Type{<:AbstractTimeSeries}) = true
-Tables.rowaccess(::Type{<:TimeArray}) = true
-Tables.rows(ta::TimeArray) = Tables.rows(Tables.columntable(ta))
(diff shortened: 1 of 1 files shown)


iblis17:
The reason that we defined `schema` is for the time column.
Since the fallback function return only the data column. for example, `[:a, :b, :c]`, but we defined the custom `schema` to return `[:timestamp, :a, :b, :c]`

quinnj:
> The reason that we defined `schema` is for the time column. Since the fallback function return only the data column. for example, `[:a, :b, :c]`, but we defined the custom `schema` to return `[:timestamp, :a, :b, :c]`

Yep, that's perfectly fine/valid.

PR #505, opened 2022-02-22
Remove candletick series labels

nilshg:
The current recipe produces a separate label for each bar, which completely clutters the plot if a user calls `plot(ta; legend = :topleft)` (or any other legend position). With this change no labels are produced by default, allowing the user to combine the candlestick plot with other plots and add a legend without having dozens of `y1, y2, y3`... entries in the legend.


nilshg:
Bump

iblislin:

PR #510, opened 2022-04-06
add compatibility with vscode tableviewer

dwinrow:
The [vscode tableviewer guide](https://www.julia-vscode.org/docs/latest/userguide/grid/) is incorrect when it says it works with all [Tables.jl](https://tables.juliadata.org/stable/#Implementing-the-Interface-(i.e.-becoming-a-Tables.jl-source)) compatible types.  In fact it requires the type's iterator to be a named tuple and uses two further packages for this.

It uses [TableTraits.jl](https://github.com/queryverse/TableTraits.jl) isiterabletable(x) to determine if the table x is iterable.  This checks the eltype of the table x requiring it to be a NamedTuple.
It also uses [IteratorInterfaceExtensions.jl](https://github.com/queryverse/IteratorInterfaceExtensions.jl) getiterator(x) to retrieve the iterator for the interface.

This commit implements a fix for #508 in the way that [DataFrames.jl implements tables.jl](https://github.com/JuliaData/DataFrames.jl/blob/main/src/other/tables.jl) by overloading the above methods.  I've also overloaded IteratorInterfaceExtensions.isiterable(x) but this may be unnecessary.


dwinrow:
Could I have some feedback on this?  I've resolved the conflicts in Project.toml, and this commit enhances the use of TimeArrays in VSCode as you can use the table viewer in the Julia extension.

iblislin:

Issue #517, opened 2023-05-12
typos


spaette:
@iblislin 

_per request I can submit a pull_

Exchange
explicitly
widths
wrappers

```
$ grep -nr Exchage TimeSeries.jl
TimeSeries.jl/docs/src/combine.md:20:down the New York Stock Exchage. Apple Computer trades on the electronic
$ grep -nr explicity TimeSeries.jl
TimeSeries.jl/NEWS.md:395:* test/combine.jl and test/split.jl now imports Base.Dates explicity
$ grep -nr withs TimeSeries.jl
TimeSeries.jl/src/timearray.jl:238:    # calculate column withs
$ grep -nr wrapers TimeSeries.jl
TimeSeries.jl/src/timearray.jl:471:###### element wrapers ###########
$ 
```

PR #518, opened 2023-05-15
fix typos

spaette:
https://github.com/JuliaStats/TimeSeries.jl/issues/517

NEWS.md
docs/src/combine.md
src/timearray.jl


iblislin:
Thanks for your contributions

PR #520, opened 2023-11-15
Changes to PrettyTables printing

### LINKED ISSUE #428: TimeArray printing return error in too small terminal
Hello,
if the terminal window is too small (e.g. VSCode) trying to print a TimeArray results in an obscure error.

> ERROR: BoundsError: attempt to access 0-element view(::Array{DateTime,1}, Int64[]) with eltype DateTime at index [1]

### LINKED ISSUE #514: Really inefficient printing for large TimeArrays
The main source seems to be this line:

https://github.com/JuliaStats/TimeSeries.jl/blob/0356f4bbad7e7f7b6a8cacde04f0948a65fd4ed4/src/timearray.jl#L297

only a very small subset of `strs` is used later.

ValentinKaisermayer:
Gets rid of a lot of code so that we do not have to reinvent the wheel. Printing is a little different (tail is omitted, top and bottom hline). 

closes #514 
closes #428


ValentinKaisermayer:
@iblislin Can you review this?

src/TimeSeries.jl
```diff
@@ -9,6 +9,7 @@ using DocStringExtensions: SIGNATURES
 using RecipesBase
 using Reexport
 using Tables
+import PrettyTables
```
iblislin, line 12: Is it fine to use `using`?

ValentinKaisermayer: Sure

iblislin:

PR #523, opened 2023-12-03
changes to vcrop = middle

ValentinKaisermayer:


ValentinKaisermayer:
More similar to the old behaviour

PR #525, opened 2024-04-22
Fixdocs

dwinrow:
The documenter makefile appears to be failing because of `ORCA`, a deprecated package which enables `savefig`.  This function is now included in `Plots`.


iblislin:
:+1:

PR #528, opened 2024-07-17
Adds method for splitting data into periods

ValentinKaisermayer:


src/split.jl
```diff
@@ -58,3 +58,37 @@ end
...
+
+Split `data` by `period` function, returns a vector of `TimeSeries.TimeArray`.
+
+## Arguments
+
+- `data::TimeSeries.TimeArray`: Data to split
+- `period::Function`: Function, e.g. `Dates.day` that is used to split the `data`.
+"""
+split(data::TimeSeries.TimeArray, period::Function) = map(i -> data[i], _split(data, period))
+function _split(data::TimeSeries.TimeArray, period::Function)
```
iblislin, line 74: well, I'm think about : does return an iterator better than just an Array?

ValentinKaisermayer: I now return an `Iterators.map`

src/split.jl
```diff
@@ -58,3 +58,35 @@ end
...
+            t1 = period(ts[i+1])
+            t0 == t1 && continue
+            push!(idx, j:i)
+            j = i + 1
+            t0 = t1
+        end
+        push!(idx, j:m)
+    end
+    return Iterators.map(i -> ts[i], idx)
+end
```
iblislin, line 92: add trialing new line of file
```suggestion
end

```

test/split.jl
```diff
@@ -1,121 +1,129 @@
...
-
 using MarketData
-
 using TimeSeries
 
-
 @testset "split" begin
 
-
-@testset "find methods" begin
```
iblislin, line 12: well, could you send the formatting patch as another PR? 
(and maybe introduce some formatter in the standalone PR, e.g. https://github.com/julia-actions/julia-format)

This cause the patch not human readable.

ValentinKaisermayer: #530

ValentinKaisermayer:
> Is this included in the #529 ?

Yes

PR #530, opened 2024-12-16
Adds formatting workflow

ValentinKaisermayer:


ValentinKaisermayer:
cc @iblislin

ValentinKaisermayer:
There are many possibilities here.

This creates a PR with the formatted code. 
Another possibility would be to check if the code in the PR is formatted. This would have the benefit that the history is cleaner and blame would show the author and not the formatting PR.

iblislin:
ah, I will vote for the later option.

ValentinKaisermayer:
Formatting check failed as expected.

ValentinKaisermayer:
I have no idea why the check still fails. If formatted the files locally and pushed them. 

```julia
julia> format(".")
true
```

ValentinKaisermayer:
@iblislin should be ready now

PR #532, opened 2025-01-19
Removes docs on dotfile

### LINKED ISSUE #531: Remove doc section about customized printing
Since `PrettyTables.jl` is used, this is no longer valid.

ValentinKaisermayer:
Closes #531


ValentinKaisermayer:
@iblislin

iblislin:
well, I think we should add some deprecate warning first.
Then, scheduling for removal.

Maybe add a banner in Documenter:
https://documenter.juliadocs.org/stable/showcase/#Warning-admonition

ValentinKaisermayer:
I don't think so. The breaking change has already happened with v0.24, so this is more like a patch.

iblislin:
ah, ok.

could you rebase this PR?

ValentinKaisermayer:
I rebased it

PR #540, opened 2025-11-11
hc/update_PrettyTables

hannahchubin:
I updated PrettyTables to version 3.1.0 and all the tests in `test/runtests.jl` are passing. I am trying to update PrettyTables in some other repos that also depend on TimeSeries, which is currently limiting the PrettyTables version to less than `v3.0.0`.


iblislin:
hmm any idea about the CI failure?

briederer:
#541 is only an issue to reactivate the github workflow again, that PRs like yours do not need to be made manually in the future (see #542).

I has a quick look at the CI output ofnyour PR and master and they are different.
The one on master should be actually fixed now since it was only to a missing compat in MarketData.jl (see JuliaQuant/MarketData.jl#105).

For your PR it seems that the `_text__print_table(...)` signature hast changed from PrettyTables v2 to v3. This needs to be fixed. I can have a quick look if you want.

PR #547, opened 2025-12-29
Added docstrings to exported functions!

anurag-mds:
This PR addresses issue #533  by ensuring that all exported methods in TimeSeries.jl have comprehensive docstrings.

The following actions were taken:

Reviewed all exported functions, types, and constants.
Added or improved docstrings for any missing or incomplete documentation.
Followed the existing documentation style for consistency.
Included examples where helpful.

Motivation:
Having complete docstrings for all exported methods improves usability and discoverability for users, making the package easier to use and understand.
Notes:
All existing tests are passing. No change was done to code just documentation-only fixes were done.
Reviewers:
@ValentinKaisermayer @iblislin

PR #549, opened 2025-12-31
Share doc strings branch

anurag-mds:
This PR implements the approach discussed in issue #547 by moving shared docstrings from exported functions into the docs directory and referencing them via @docs, eliminating duplication while preserving rendered documentation.

@iblislin Please review.


anurag-mds:
Thanks for checking. 
My understanding is that methods extending Base usually aren't listed explicitly in the public API, and internal helpers(_ : prefixed)
aren't part of it either

That said, I'm happy to follow whatever convention timeseries.jl prefers 
Just do let me know about it

ValentinKaisermayer:
https://documenter.juliadocs.org/stable/man/syntax/#Info-67d1dac5cc53b84b

Documenter.jl should only consider exported symbols and those marked with `public`. So I wonder why we export `Base.+` ?

ValentinKaisermayer:
> https://documenter.juliadocs.org/stable/man/syntax/#Info-67d1dac5cc53b84b
> 
> Documenter.jl should only consider exported symbols and those marked with `public`. So I wonder why we export `Base.+` ? 

Ah, sry got that wrong. We document it, i.e. have a docstring but they are not exported so `@autodocs` complains.

ValentinKaisermayer:
DataFrames.jl simply documents everything, including, e.g. `Base.hcat`. Maybe this is the right approach. 

https://dataframes.juliadata.org/stable/lib/functions/#Base.hcat

anurag-mds:
I see the trade-off.
My intial assumption was to exclude base extensions and internal helpers from public API, but dataframes.jl documenting Base methods is a strong precedent.

I'm fine adopting that approach to maintain consistency.


To confirm before I proceed:
Should I include the reported Base methods in the API section while keeping internal helpers prefixed with _ excluded?

iblislin:
I vote for including all `Base.*` methods in the API reference page, like DataFrames.jl does.

@anurag-mds could you try out the `public` keyword for `Base` methods?

>https://documenter.juliadocs.org/stable/man/syntax/#Info-67d1dac5cc53b84b
>
> Documenter.jl should only consider exported symbols and those marked with public. So I wonder why we export Base.+ ?

anurag-mds:
Sounds good
I’ll mark the relevant Base.* methods as public and update the API reference accordingly.
I’ll report back if I hit any issues with Documenter

anurag-mds:
Apologies for the delay on this update - I’m currently away from my development setup for a few days, but I’ve noted everything and will push the changes as soon as I’m back. I’ll update here if I run into any issues with Documenter.

Thanks for the guidance and patience

anurag-mds:
@iblislin I have done all the necessary changes Kindly review it and any changes do let me know!

Project.toml
```diff
@@ -22,7 +22,7 @@ RecipesBase = "0.5, 0.7, 0.8, 1.0"
 Reexport = "1"
 Statistics = "1"
 Tables = "1"
-julia = "1.6"
+julia = "1.11"
```
iblislin: Is it necessary to change the version of Julia?
or we could switch to the LTS version, `1.10` I think.

anurag-mds:
Does this seems good?

iblislin:

iblislin:
The new ref page looks fine: https://juliastats.org/TimeSeries.jl/dev/api/.
I'm going to make a new release.

anurag-mds:
Thanks a lot for the review and for merging this!
I really appreciate the guidance throughout and learned a lot from the process. Glad the new API page looks good.

This PR helped me learn how to rebase, understand reviews, and make the necessary modifications.

PR #553, opened 2026-03-10
Add time_slots and TimeWindow for issue #481

### LINKED ISSUE #481: Time slots iterators and helper functions
- I need a iterator to create continuous time slots in a given interval of a day.
    - e.g. start from 2021/1/1 to 2021/1/10, in each day, I need to create 15m time slots from am 8:45 to pm 1:45.
- Helper functions:
    - Given a timestamp, find the time slots which the timestamp belongs to.

anurag-mds:
Implements the API proposed in #481.

`time_slots(start, stop, interval; from, to)` — generates a `TimeArray` of timestamps within a daily time window
`time_slots(range; from, to)` — same, accepts a `StepRange{DateTime}` directly
`when(ta, time_slots(; from, to))` — filters an existing `TimeArray` by time-of-day via the new `TimeWindow` type

Closes #481


src/split.jl
```diff
@@ -10,6 +10,57 @@ function when(ta::TimeArray, period::Function, t::Integer)
...
+
+time_slots(; from::Time, to::Time) = TimeWindow(from, to)
+
+function when(ta::TimeArray, window::TimeWindow)
+    return ta[findall(t -> window.from <= Time(t) <= window.to, timestamp(ta))]
+end
+
+"""
+    time_slots(start::DateTime, stop::DateTime, interval::Period;
+                    from::Time=Time(0,0), to::Time=Time(23,59))
```
iblislin:
```suggestion
    time_slots(start::DateTime, stop::DateTime, interval::Period;
               from::Time=Time(0, 0), to::Time=Time(23, 59))
```

iblislin: is that original indentation suggested by the formatter or documenter?

src/split.jl
```diff
@@ -10,6 +10,57 @@ function when(ta::TimeArray, period::Function, t::Integer)
...
+"""
+function time_slots(
+    start::DateTime,
+    stop::DateTime,
+    interval::Period;
+    from::Time=Time(0, 0),
+    to::Time=Time(23, 59),
+)
+    timestamps = filter(dt -> Time(dt) ∈ from:interval:to, start:interval:stop)
+    return TimeArray(timestamps, ones(length(timestamps)))
```
iblislin: oh, could the design be like: returning a lazy iterator that do not materialized the vector at first?

since I'm not sure returns with `ones` array is a good idea or not.

I just think an iterator will be fine to use in TimeArray constructor?

anurag-mds: You're right returning ones as dummy values doesn't make sense here. The meaningful output is just the timestamps themselves. I can change `time_slots` to return `Vector{DateTime}` directly and let the user wrap it in a `TimeArray` with their own values. Does that direction make sense?

iblislin: Ah, the direction is right, and maybe we could take one step forward:
Make `time_slots` to return something like `TimeSlot{DateTime}` .
This `TimeSlot` is an  iterator type that behaves like `Range` type( recall that: `Date(...):Date(...)` also creates a iterator as well).
Let user materialize it via `collect` on demand.

anurag-mds:
@iblislin

src/split.jl
```diff
@@ -10,6 +10,90 @@ function when(ta::TimeArray, period::Function, t::Integer)
 end
 when(ta::TimeArray, period::Function, t::String) = ta[findall(period.(timestamp(ta)) .== t)]
 
+"""
+    when(ta::TimeArray, window::TimeWindow)
+
+Return a subset of `TimeArray` where the time of each timestamp falls
+within the daily window `[window.from, window.to]`.
+"""
```
iblislin: ah, please re-locate this docstring

src/split.jl
```diff
@@ -10,6 +10,90 @@ function when(ta::TimeArray, period::Function, t::Integer)
...
+Base.IteratorSize(::Type{<:TimeSlot}) = Base.SizeUnknown()
+Base.eltype(::Type{<:TimeSlot}) = DateTime
+
+function Base.iterate(ts::TimeSlot)
+    current = ts.start
+    while current <= ts.stop
+        if ts.from <= Time(current) <= ts.to
+            return (current, current + ts.interval)
+        end
+        current += ts.interval
```
iblislin, line 47: what if the user set ... an negative Period? e.g. `Minute(-1)` ?

anurag-mds: I completely missed that point.
Thanks!

src/split.jl
anurag-mds: Maybe while formatting I misplaced it

anurag-mds:
I forgot to mention I added the following tests to make sure the negative inputs are handled
`@test_throws ArgumentError time_slots(now(), now()+Day(1), Minute(-1))`

Error thrown:
```julia
julia> time_slots(now(), now()+Day(1), Minute(-1))
ERROR: ArgumentError: interval must be positive, got -1 minute
Stacktrace:
 [1] #time_slots#25
   @ C:\Users\RadhaKunjBalram\Documents\open source julia\TimeSeries.jl\src\split.jl:80 [inlined]
 [2] time_slots(start::DateTime, stop::DateTime, interval::Minute)
   @ TimeSeries C:\Users\RadhaKunjBalram\Documents\open source julia\TimeSeries.jl\src\split.jl:71
 [3] top-level scope
   @ REPL[4]:1
```

```julia
julia> time_slots(
               DateTime(2023, 1, 1), DateTime(2023, 1, 2), Day(-1), from=Time(9), to=Time(17)
           )
ERROR: ArgumentError: interval must be positive, got -1 day
Stacktrace:
 [1] #time_slots#25
   @ C:\Users\RadhaKunjBalram\Documents\open source julia\TimeSeries.jl\src\split.jl:80 [inlined]
 [2] top-level scope
   @ REPL[2]:1

julia> 
```

```julia
julia> rng = DateTime(2023, 1, 2):Day(-1):DateTime(2023, 1, 1)
DateTime("2023-01-02T00:00:00"):Day(-1):DateTime("2023-01-01T00:00:00")

julia> time_slots(rng)
ERROR: ArgumentError: interval must be positive, got -1 day
Stacktrace:
 [1] #time_slots#26
   @ C:\Users\RadhaKunjBalram\Documents\open source julia\TimeSeries.jl\src\split.jl:98 [inlined]
 [2] time_slots(range::StepRange{DateTime, Day})
   @ TimeSeries C:\Users\RadhaKunjBalram\Documents\open source julia\TimeSeries.jl\src\split.jl:94
 [3] top-level scope
   @ REPL[6]:1

julia> 
```

src/split.jl
```diff
@@ -10,6 +10,97 @@ function when(ta::TimeArray, period::Function, t::Integer)
...
+    return ta[findall(t -> window.from <= Time(t) <= window.to, timestamp(ta))]
+end
+
+struct TimeSlot{T<:DateTime,P<:Period}
+    start::DateTime
+    stop::DateTime
+    interval::P
+    from::Time
+    to::Time
+end
```
iblislin: I got you want this?
```suggestion
struct TimeSlot{T<:DateTime,P<:Period}
    start::T
    stop::T
    interval::P
    from::Time
    to::Time
end
```

anurag-mds: Done,
I changed the docs as the functions now return a `TimeSlot` iterator and parameterize `time_slots` and `TimeSlot` for generic `DateTime` types.

src/split.jl
```diff
@@ -10,6 +10,93 @@ function when(ta::TimeArray, period::Function, t::Integer)
...
+    time_slots(start::T, stop::T, interval::Period;
+                    from::Time=Time(0,0), to::Time=Time(23,59))
+
+Return A `TimeSlot` iterator where each timestamp falls within the daily time 
+window `[from, to]`,
+filtered from the full `start` to `stop` range at the given `interval`
+"""
+function time_slots(
+    start::T, stop::T, interval::Period; from::Time=Time(0, 0), to::Time=Time(23, 59)
+) where {T<:TimeType}
```
iblislin:
```suggestion
function time_slots(
    start::T, stop::T, interval::P; from::Time=Time(0, 0), to::Time=Time(23, 59)
) where {T<:TimeType,P<:Period}
```

then, also use those type variables in return?

src/TimeSeries.jl
```diff
@@ -14,6 +14,8 @@ using PrettyTables: PrettyTables, pretty_table
 export TimeArray,
     AbstractTimeSeries,
     when,
+    time_slots,
+    TimeWindow,
```
iblislin, line 18: ah, consider export `TimeSlot` as well?

src/split.jl
```diff
@@ -10,6 +10,93 @@ function when(ta::TimeArray, period::Function, t::Integer)
...
+"""
+function time_slots(
+    range::StepRange{T}; from::Time=Time(0, 0), to::Time=Time(23, 59)
+) where {T}
+    if step(range) <= zero(step(range))
+        throw(ArgumentError("interval must be positive, got $(step(range))"))
+    end
+    return TimeSlot{eltype(range),typeof(step(range))}(
+        range.start, range.stop, step(range), from, to
+    )
```
iblislin: I'm not sure about the current behaviour of compiler, is this consider type stable or not, if we use eltype(..) and typeof(..) in return statement. 🤔 

type stable is preferred, still fine if not.

iblislin:
I'm still thinking about the possibility to support negative interval,
especially an edge case like: `DateTime(2023, 1, 2, 0, 3):Minute(-4):DateTime(2023, 1, 1)`.

> I forgot to mention I added the following tests to make sure the negative inputs are handled `@test_throws ArgumentError time_slots(now(), now()+Day(1), Minute(-1))`
> 
> Error thrown:
> 
> ```julia
> julia> time_slots(now(), now()+Day(1), Minute(-1))
> ERROR: ArgumentError: interval must be positive, got -1 minute
> Stacktrace:
>  [1] #time_slots#25
>    @ C:\Users\RadhaKunjBalram\Documents\open source julia\TimeSeries.jl\src\split.jl:80 [inlined]
>  [2] time_slots(start::DateTime, stop::DateTime, interval::Minute)
>    @ TimeSeries C:\Users\RadhaKunjBalram\Documents\open source julia\TimeSeries.jl\src\split.jl:71
>  [3] top-level scope
>    @ REPL[4]:1
> ```
> 
> ```julia
> julia> time_slots(
>                DateTime(2023, 1, 1), DateTime(2023, 1, 2), Day(-1), from=Time(9), to=Time(17)
>            )
> ERROR: ArgumentError: interval must be positive, got -1 day
> Stacktrace:
>  [1] #time_slots#25
>    @ C:\Users\RadhaKunjBalram\Documents\open source julia\TimeSeries.jl\src\split.jl:80 [inlined]
>  [2] top-level scope
>    @ REPL[2]:1
> 
> julia> 
> ```
> 
> ```julia
> julia> rng = DateTime(2023, 1, 2):Day(-1):DateTime(2023, 1, 1)
> DateTime("2023-01-02T00:00:00"):Day(-1):DateTime("2023-01-01T00:00:00")
> 
> julia> time_slots(rng)
> ERROR: ArgumentError: interval must be positive, got -1 day
> Stacktrace:
>  [1] #time_slots#26
>    @ C:\Users\RadhaKunjBalram\Documents\open source julia\TimeSeries.jl\src\split.jl:98 [inlined]
>  [2] time_slots(range::StepRange{DateTime, Day})
>    @ TimeSeries C:\Users\RadhaKunjBalram\Documents\open source julia\TimeSeries.jl\src\split.jl:94
>  [3] top-level scope
>    @ REPL[6]:1
> 
> julia> 
> ```

iblislin:
ah, I think we could keep current implementation for the negative interval.
Could you add some notes in the code (maybe a TODO or FIXME section as comment) about that negative interval? (also note a edge case: `DateTime(2023, 1, 2, 0, 3):Minute(-4):DateTime(2023, 1, 1)`)
maybe we could relax this non-negative constrain later.

anurag-mds:
> ah, I think we could keep current implementation for the negative interval. Could you add some notes in the code (maybe a TODO or FIXME section as comment) about that negative interval? (also note a edge case: `DateTime(2023, 1, 2, 0, 3):Minute(-4):DateTime(2023, 1, 1)`) maybe we could relax this non-negative constrain later.

Sure!, Let me do this real quick.

iblislin:

